Issue #22, opened 2019-04-05
Feature Request: reset 'still watching' state


jonesthefox:
Hey there, thanks for your good work, it's a really useful addon. 

Is it possible to reset the 'still watching' state after stopping the video when the 'still watching' button is not pressed?

Let's say, i watch i.e. 5 episodes and i was too lame to press that button, i then select the next episode and have to press the 'still watching' button when the episode ends, that i can watch the next 5 episodes without touching my remote. 
The behauvior should be, that it plays my 5 episodes and ask me then 'are you still there?'..

im85288:
Should be resolved with the above PR

Issue #20, opened 2019-03-02
Media Stop During Up Next Popup


tmortonorr:
If you stop the currently playing media while the up next popup is showing (without pressing the close button on the popup) it stops the media but then plays the next episode anyway.

im85288:
Should be resolved with the above PR

PR #30, opened 2019-05-11
Reset last_file and data upon user stopping playback

angelblue05:
This fixes the scenario where user starts playing a file, stops before last_file gets updated in up next on the next episode. Then users plays the same file, leaving last_file outdated and preventing the dialog from showing up again. 

While the up next dialog is displayed, this will prevent a user manually stopping playback from triggering up next, data will be cleared.


angelblue05:
Should fix this PR 
https://github.com/im85288/service.upnext/issues/20

angelblue05:
It should also fix
https://github.com/im85288/service.upnext/issues/22

im85288:
Awesome thanks Tammy

PR #31, opened 2019-05-11
removed tabs in resources/lib/player.py

jonesthefox:
removed tabs, because it broke the addon so kodi won't start it.


angelblue05:
Hmm, sorry about that. I had not noticed, my ide must be messed up somehow.

im85288:
Thanks guys

Issue #18, opened 2019-02-15
Next Episode popup not Showing


agentxan:
Libreelec 9
KODI 18
Upnext 1.0.0
Embycon 1.8.12

I am using embycon and upnext. After the first episode plays upnext pops up and goes to to the second episode but after the second episode it does not pop up for the third.

I added a log line to reset_addon_data() and sometimes it looks like it is being called after embycon sends the next episode data.

[kodi2.zip](https://github.com/im85288/service.upnext/files/2870083/kodi2.zip)

angelblue05:
This should be resolved with up next version 1.0.1

agentxan:
I have been using the latest from github and the problem looks to be fixed.

ShaVerZ:
The popup doesnt even appear for me, Seren and Venom. Tried all versions of Up Next, inc. v1.0.7

Hopefully a fix is in the works.....

im85288:
> The popup doesnt even appear for me, Seren and Venom. Tried all versions of Up Next, inc. v1.0.7
> 
> Hopefully a fix is in the works.....

Open a new issue and follow the wiki on how to report a bug. Without that information nobody can guess what’s causing the issue on your system.

123Venom:
I noticed the same issue...no version after 1.0.1 displays the notification anymore.

ShaVerZ:
Thanks for your reply, I'll report the issue today.
I didnt report an issue previously because I was told (Reddit) that you are aware of the issues and a fix was in the works.....

123Venom:
> 
> 
> Thanks for your reply, I'll report the issue today.
> I didnt report an issue previously because I was told (Reddit) that you are aware of the issues and a fix was in the works.....

I'm not convince the issue is the addons.  Both Venom and Seren work fine with 1.0.1 but the moment you update beyond that version neither addon produces the pop up notification anymore.  Best way to tell is for me to log the data and see what's going on and I've not had a chance to yet.  I thought I read that 1.0.1 was the last version that used Addon Signals?  If so how does the communication then take place?  Is an addon change required for that?

ShaVerZ:
> communication

@123Venom which version of Kodi are you using, my problems started when I updated Kodi to v18

123Venom:
[...]

Issue #17, opened 2019-02-08
OSMC (Leia) - Audio delay and only playing one item


danmed:
i didnt turn debugging on until about 18:00 so this log contains the issue both with and without debugging on..

https://paste.kodi.tv/upekiteyun.kodi

I have this issue on both LibreElec and OSMC (both Leia)

It will play 1 extra episode, but not a 2nd.. Equally in OSMC, the audio plays for 10 seconds whilst the video is frozen, then the video plays with no audio until the sync up.. This doesn't happen on LibreElec.

The other issue is that this does not work at all with Smart Playlists.. it finds the next episode and counts down, but it never plays it (This is not in the debug log.. but hopefully you'll know what's up)..

This is running off a MySQL DB btw.

danmed:
OK, i think i've fixed this with a fresh install of OSMC.. the audio delay problem is there but i don't think thats your problem as it's not present in LibreElec.

angelblue05:
@danmed 
I know it's been a little while, but is your issue resolved with UpNext version 1.0.1?

danmed:
I'm afraid I havn't used it since.. I just set Kodi to auto play the next
episode until I tell it to stop.

angelblue05:
Ok thanks for replying back. Well if you want to give it another try someday. I am fairly certain this has been resolved.

Issue #11, opened 2018-12-22
Instant show again after play next


hawkeyexp:
If using play next at the end of a show new episode starts and again a play next is shown for next episode. Sems to be a problem on kodi side while triggerig more than one event - possible you can add a fix for it to ignore new events within 5 seconds etc.

angus2064:
Same thing for me ,regardless the skin used . 
I use Kodi 18RC3 on a Shield TV 2018 (test on RC4 ,same bug).

im85288:
Hmm not seeing this, let’s wait for Kodi 18 to be finalised before checking this again

Mecool-KX:
Fist of all, thank you for your great work

I have the same problem with Kodi 18 RC4

ryanjerskine:
I'm still seeing the same problem. Any details I can provide that would help get this fixed?

im85288:
Can someone upload a log? Please follow the link on the wiki of how to capture this for this add on. Thanks

ryanjerskine:
I can upload a log. I’m currently using the Google Drive plugin for all of my playback. Does that make a difference? I can switch to local playback to grab the log if I need to.

im85288:
I am not sure as I have not seen it when using either Netflix or EmbyCon as the source. Any log with it reproduced would be helpful thanks

seanrand:
I can reproduce this reliably when I click "watch next" on the "still there?" prompt after x episodes. I'll see if I can get logs.

notoco:
In the case of Netflix, this error occurs on version 0.13.21. On version 14.0b20 it is not displayed. But version 14.0b21 does not correctly indicate episodes of serials added to the library as viewed

im85288:
Personally I use the newer version 14 as it integrates better with this add on by sending a notification at the correct time to show the up next dialog (rather than waiting until 30 seconds before).

Anyway if someone can get some logs uploaded I’m sure it’ll be a simple fix

notoco:
Well, but you already have logs from two people :)
Full log - from start kodi to end.
Start episode 5... scroll to end, popup, play next and popup is show at beggining of next episode

im85288:
Whoops did not see the edited versions with logs attached. Thanks guys

notoco:
BTW - change 
`        <setting id="autoPlaySeasonTime" type="slider" label="30001" default="30" range="10,5,60" option="int" />
`
to (for example ofc)
[...]

PR #32, opened 2019-05-12
Fix dialog not showing up

angelblue05:
Having the reset in OnPlayBackStarted causes the dialog to sometimes not show up, especially if up next information is sent as OnPlayBackStarted triggers. 

It will also correct the dialog showing up when transitioning to the next episode. It discards the saved data once we are done with it.

### DIFF
--- a/resources/lib/playbackManager.py
+++ b/resources/lib/playbackManager.py
@@ -30,6 +30,7 @@ class PlaybackManager:
             return
         self.log("episode details %s" % json.dumps(episode), 2)
         self.launch_popup(episode)
+        self.api.reset_addon_data()
 
     def launch_popup(self, episode):
         episode_id = episode["episodeid"]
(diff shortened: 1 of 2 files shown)


angelblue05:
This PR should fix this
https://github.com/im85288/service.upnext/issues/18

angelblue05:
This PR might also fix this one
https://github.com/im85288/service.upnext/issues/11

angelblue05:
I also readjusted this one here, since the PR's fix was 8 spaces. It just needs 4.
https://github.com/im85288/service.upnext/pull/31

angelblue05:
It should also correct this issue, the playing only a single item after the initial one.
https://github.com/im85288/service.upnext/issues/17

im85288:
Awesome thanks

Havokvalor:
Will this change be available on the kodi repository?

im85288:
Yes it will take some time but I’ll be pushing these changes to the kodi repository

PR #33, opened 2019-05-13
Fix play next race condition

angelblue05:
If the play data is received before onPlayBackStated is called, the player state is then outdated with previous total time and play time, causing up next to trigger as play next starts.


angelblue05:
This should completely fix this one.
https://github.com/im85288/service.upnext/issues/11

im85288:
Cheers 😊

Issue #34, opened 2019-05-15
Continuous loading after continue watching


Havokvalor:
I added the changes made recently to the addon and it seems that the prompt showing up at the beginning of the episode is gone, however now sometimes you will click continue watching and it will continuously show a loading circle while the episode plays, the only way to stop it from doing this is to hit home to exit out of kodi and then go back in.

angelblue05:
Please provide your Kodi log. I have not experienced this on my side. Thanks.

angelblue05:
Does this PR helps?
https://github.com/im85288/service.upnext/pull/35

thebertster:
I'm experiencing the issue on my Raspberry Pi (LibreElec 9.0.2MR / Leia 18.2) where fairly consistently when we start playing the next episode, we immediately trigger again (I do still get the prompt, not just a spinning circle though). I do not see this (nor have I ever seen any problem) on my other Kodi box which is a faster Linux box, otherwise identically configured. It feels like this may be related to this behaviour.

From a bit of debugging, it looks like for a brief period after we trigger the next episode to play, the call to getTime() is returning the playback timer from the end of the previous episode, meaning (total_time - play_time) is typically very small (sometimes even negative) for a short period. This looks to be a Kodi bug, perhaps only manifesting on slower systems?

I have worked around this by changing the test to (5 <= total_time - play_time <= int(notification_time)).

thebertster:
After a bit more testing and debugging, it seems that occasionally the call to getTime() shortly after the playback started event is fired can actually return a 'not playing' exception, which then causes tracking to be disabled, preventing Up Next from triggering at the end of that episode.

It seems like there is a race condition where the playback started event is triggered putting the monitor in tracking enabled state, but the calls to getTime() from within the monitor loop are returning results which lag behind the actual playback state, returning either the playback timer from the previously-playing video, or an exception that the playback is not in progress.

I'll continue to try to troubleshoot since I seem to be able to reproduce the issues fairly reliably.

thebertster:
Checking whether the behaviour is related to the Kodi setting to pause playback to wait for the screen refresh rate to change to match the video frame rate...

angelblue05:
I don’t see how that is possible. Tracking should disable itself after the pop up and only start again when onPlaybackStarted hits. Are you saying that at that point, the TotalTime is still wrong in your situation?? What if you add ab xbmc.sleep(200) right before self.track = True in player.py?

thebertster:
[...]

PR #37, opened 2019-05-21
Fix race condition on slower devices

angelblue05:
Delay start tracking by 5s to ensure player information has the chance to update on slower devices and prevent up next from showing right as playback starts.

### DIFF
--- a/resources/lib/player.py
+++ b/resources/lib/player.py
@@ -29,6 +29,7 @@ class Player(xbmc.Player):
 
     def onPlayBackStarted(self):
         # Will be called when kodi starts playing a file
+        xbmc.sleep(5000) # Delay for slower devices, should really use onAVStarted for Leia
         self.track = True
         if utils.settings("developerMode") == "true":
             self.developer.developer_play_back()


angelblue05:
For the issue reported here. However the OP mentions spinning circles which I have not been able to replicate.
https://github.com/im85288/service.upnext/issues/34

im85288:
Cheers, I seen the comment about using onAVPlayback, I did try that on first implementation but it didn’t work. Maybe worth a retry in the future

angelblue05:
The only reason I didn’t make the onAVStarted change is that I think up next is also on Krypton and that is a Leia change only.

im85288:
No it’s Leia Only, I left the next up one as compatible with krypton. If it works for you by all means go for it, but for me it did not when trying it about 6 months ago (it was picked up on the kodi code review as a suggested change)

angelblue05:
Oh ok. Well I have it running on Krypton lol :)
Works great.

But if its Leia only, I’ll make the adjustment.

Issue #36, opened 2019-05-21
Automatic Short Play Not Working


zang74:
Hi Ian.

I've had this problem for a bit and hadn't had much of a chance to look at it myself (I created the original "automatic play short videos" PR), but see a lot has changed under the hood and I'm not *that* competent a python coder to grok all that's going on here now. Anyhow, I decided to try after your updates of the last few days (dled directly from GitHub today), but it still seemingly exists.

The problem:  
Having the time set and playing a video shorter than that will not automatically move to the next, regardless of whether the on screen notification is displayed. It's autoplaying (and logging) regular videos just fine. Short videos however, it just drops out of the player as it would without Up Next installed. And there's nothing in the logs to indicate why.

im85288:
Hi Zang,

There’s a check in monitor.py that does not enable the addon for any videos less than 3 minutes long. Is the short play video you mention less than that time? I could remove the check as it was just to stop it messing with trailers etc.

zang74:
Thanks, Ian.

It was actually set for 300 seconds/5 minutes. Removing that line from monitor.py locally seems to have resolved that problem.

Your trailer skipping functionality need not be lost though, as the “official” Kodi trailer naming convention is “moviename-trailer.ext”. Filtering out by the inclusion of “-trailer.[ext]” should do the same for most cases.

im85288:
Ah yes 300 seconds, been a long time since I looked at the code! I will remove that check for the next release as it probably causes more confusion than I originally anticipated. Thanks for confirming all is good.

zang74:
No problem.

I did however find that UpNext threw an error when disabling/reenabling. Upon disabling, I got a "check log files for details" error with the following in the log:

`2019-05-21 15:57:56.905 T:3688825712   ERROR: EXCEPTION: Unknown addon id 'service.upnext'.
2019-05-21 15:57:56.906 T:3688825712   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                             - NOTE: IGNORING THIS CAN LEAD TO MEMORY LEAKS!
                                            Error Type: <type 'exceptions.RuntimeError'>
                                            Error Contents: Unknown addon id 'service.upnext'.
                                            Traceback (most recent call last):
                                              File "/storage/.kodi/addons/service.upnext/service.py", line 4, in <module>
                                                Monitor().run()
                                              File "/storage/.kodi/addons/service.upnext/resources/lib/monitor.py", line 51, in run
[...]

PR #38, opened 2019-05-22
Fix up next for short videos

angelblue05:


angelblue05:
To fix this here
https://github.com/im85288/service.upnext/issues/36

Issue #21, opened 2019-03-07
Extend the time before the notification


simonbcn:
There are TV shows whose credit titles last longer than 1 minute.

notoco:
https://github.com/im85288/service.upnext/issues/11#issuecomment-462773005

simonbcn:
Yes, I've tried it and it works but the notification still shows only 60 seconds. Where can I change this behavior?

angelblue05:
This has been added with the latest release. @im85288 please close this.

im85288:
Closed thanks

PR #41, opened 2019-05-23
Extend notification time to 180

angelblue05:
Modify dialogs to hide countdown for time remaining above 60 seconds. 
Add a pause/resume for the progress reported in dialogs.


angelblue05:
For feature requested here too
https://github.com/im85288/service.upnext/issues/21

Issue #24, opened 2019-04-14
Feature Request for the next episode playing at the same tempo as the current


Dnkhatri:
Sometimes I watch stuff at a faster tempo but currently have to manually change the tempo after each episode so the usefulness of the plugin decreases for me I want the upnext plugin to check the playback tempo and keep it for the next episode

angelblue05:
I honestly have no idea what you are referring to? Why should up next be in charge of this behavior? And how do I setup something like that?

Dnkhatri:
> I honestly have no idea what you are referring to? Why should up next be in charge of this behavior? And how do I setup something like that?

You have to enable sync playback to display once you do that you do that you can change the tempo the video is played at in kodi by pressing alt+right or left arrow. If I am playing an episode at 1.5x speed once the episode finishes and upnext starts the new episode the new episode starts at normal playback speed where as I would like it to keep playing at the same speed as the previous episode.

angelblue05:
I believe this should be a FR for Kodi itself. Unless you tell me that without using up next, but using the regular kodi play next automatically, Kodi keeps the speed from the first video?

@im85288 your opinion on this is needed. It is your project.

im85288:
Yes agreed, not something this addon will handle

Dnkhatri:
Okay I checked when I don't press the play next button the tempo for the next episode stays at @1.5 . But if play next it doesn't. So kodi does play next doesn't. @angelblue05 @im85288  

> I believe this should be a FR for Kodi itself. Unless you tell me that without using up next, but using the regular kodi play next automatically, Kodi keeps the speed from the first episode?
> 
> @im85288 your opinion on this is needed. It is your project.

angelblue05:
@Dnkhatri did you get a chance to try the PR to confirm this is indeed resolved now?

Dnkhatri:
Ah sorry pressed close issue  wrongly as far as I can tell  no difference if I don't press the play next button the tempo is maintained if do the new episodes starts a 1 tempo

angelblue05:
Where does the content come from? Is it from Kodi or some addon? Does it make a difference if you also enable play next automatically in the Kodi settings?

Dnkhatri:
> Where does the content come from? Is it from Kodi or some addon? Does it make a difference if you also enable play next automatically in the Kodi settings?

The content is from kodi my kodi play next was off so I tried with it on made no difference.

angelblue05:
@Dnkhatri 
[...]

PR #46, opened 2019-07-22
Add priority to existing playlist & endtime prop

angelblue05:
This is to resolve conflict between shuffled playlist and up next. If episodes are already queued up and we are not at the end of the playlist, grab the next episode to play from the existing playlist.

It also adds the endtime as a dialog property to reflect the endtime of the next episode when the dialog is displayed.

"runtime" could be added as an optional field to next_info data, to fill endtime value. The runtime should be in seconds.

### DIFF
--- a/resources/lib/api.py
+++ b/resources/lib/api.py
@@ -28,6 +28,22 @@ class Api:
     def play_kodi_item(episode):
         utils.JSONRPC("Player.Open", id=0).execute({"item": {"episodeid": episode["episodeid"]}})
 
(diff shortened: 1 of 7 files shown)


angelblue05:
This may resolve this issue by giving the existing playlist priority. If the episodes are queued up in advanced, it will not relaunch playback, but instead seek to the next episode. The tempo should hopefully be maintained.
https://github.com/im85288/service.upnext/issues/24

sualfred:
@im85288 
Tested. Works as it should.

PR #54, opened 2019-10-18
Fix prepended spaces in description

dagwieers:
The addon.xml includes indented lines in a single paragraph.
Unfortunately this causes newlines and added whitespace in the Kodi
add-on interface.

This PR fixes that.

This PR also removes trailing spaces in README.md.


im85288:
Thank you for the fixes

PR #55, opened 2019-10-18
This adds sanity checks for Python

dagwieers:
This adds sanity checks for Python, but doesn't include any fixes just
yet. Once we have everything fixed to a certain standard, we can enable
this in Travis for new contributions.

This also add a `make zip` target for creating test builds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,8 @@
-
+# Test caches
+*.pyc
 *.pyo
+.tox/
+
+# Logs
+kodi-addon-checker-report.log
+kodi-addon-checker.log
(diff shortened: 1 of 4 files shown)


dagwieers:
@im85288 I don't know if you are into this. And I'd like to also propose the required changes for PEP8 and other best practices. Maybe in a separate PR ? The included pylintrc will probably be tweaked to specific needs.

dagwieers:
It would be nice if we can enabled this in Travis and then test it for both Python 2.7, 3.6, 3.7 and 3.8 (as we do for other Kodi projects).

im85288:
Yes more than happy to have improvements whenever we can.

PR #56, opened 2019-10-18
Fix all existing flake8 issues

dagwieers:
This implements fixes for all reported flake8 issues.

Beware of two things that require more testing:
- defaulting to unicode for string literals
- unicode_to_ascii implementation


dagwieers:
I need this to be merged before I can implement all pylint fixes. (As there's some overlap)

PR #57, opened 2019-10-18
Fix all pylint issues

dagwieers:
This implements fixes for all reported flake8 issues.

Beware of two things that require more testing:
- pylint complains about that Control does not have setPercent ?

### DIFF
--- a/resources/lib/api.py
+++ b/resources/lib/api.py
@@ -2,8 +2,8 @@
 # GNU General Public License v2.0 (see COPYING or https://www.gnu.org/licenses/gpl-2.0.txt)
 
 from __future__ import absolute_import, division, unicode_literals
-import xbmc
 import json
+import xbmc
 import resources.lib.utils as utils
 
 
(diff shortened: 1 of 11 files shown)


dagwieers:
So I already prepared this, but I now have to resolve the remaining conflicts ;-)

im85288:
Cool, I’ll check back later or maybe tomorrow as it’s movie time! Cheers

dagwieers:
There is one more thing to do after this:
- Add integration and unit tests

And more importantly, unicode-related tests, because that is going to tell if it works fine for both Python 2 and Python 3. And then it would be nice if we could start the service, and signal things to test all code-paths.

We could then also add coverage support, to see if we are testing all codepaths properly. For this project a near 100% coverage is possible.

im85288:
Excellent, really appreciate the improvements and fully agree if we have tests we can be more confident of any further changes that we have not broken any of the current functionality

PR #61, opened 2019-10-19
Fix exception when exiting add-on

dagwieers:
This PR includes:
- A fix for an exception when logging when the add-on is disabled
- Replace all dict access using the more forgiving get()
- Some more unicode fixes

### DIFF
--- a/resources/lib/api.py
+++ b/resources/lib/api.py
@@ -16,7 +16,7 @@ class Api:
 
     def log(self, msg, lvl=2):
         class_name = self.__class__.__name__
-        utils.log("[%s] %s" % (utils.addon_id(), class_name), msg, int(lvl))
+        utils.log('[%s] %s' % (utils.ADDON_ID, class_name), msg, int(lvl))
 
     def has_addon_data(self):
         return self.data
@@ -30,7 +30,7 @@ class Api:
 
     @staticmethod
     def play_kodi_item(episode):
-        utils.JSONRPC("Player.Open", id=0).execute({"item": {"episodeid": episode["episodeid"]}})
+        utils.JSONRPC("Player.Open", id=0).execute({"item": {"episodeid": episode.get('episodeid')}})
 
     def get_next_in_playlist(self, position):
         result = utils.JSONRPC("Playlist.GetItems").execute({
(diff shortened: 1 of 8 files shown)


dagwieers:
@mediaminister Can you review this wrt. unicode?

dagwieers:
So wrt. Codecov, normally it should pick up *.github/codecov.yml* which disables the above reports and also ignores *test/*. It also appears that while I run the service using coverage, because it gets killed (to stop it) the coverage report is not used :-(

So I need to find a way to stop it cleanly without killing/breaking it, possibly by using a wrapper script. (I can't force a KeyboardInterrupt by using SIGINT :-/)

resources/lib/monitor.py
```diff
@@ -60,6 +61,6 @@ def onNotification(self, sender, method, data):  # pylint: disable=invalid-name
             return
 
         data = utils.decode_data(data)
-        data['id'] = "%s_play_action" % str(sender.replace(".SIGNAL", ""))
+        data['id'] = "%s_play_action" % from_unicode(sender.replace(".SIGNAL", ""))
```
mediaminister: `"%s_play_action" `is again a unicode literal because you imported `unicode_literals`. Because sender is a variable from ` xbmc.Monitor.onNotification` it makes sense here to decode `sender` to unicode, so use `to_unicode`

resources/lib/playitem.py
```diff
@@ -63,7 +64,7 @@ def handle_now_playing_result(self, result):
 
                 # Get current episodeid
                 current_episode_id = self.api.get_episode_id(
-                    showid=str(self.state.tv_show_id), show_season=current_season_id,
+                    showid=from_unicode(self.state.tv_show_id), show_season=current_season_id,
```
mediaminister: Again, because you imported `unicode_literals` we assume every string is unicode. We only decode and encode when inputting or outputting data from and to xbmc api on the edges of our program.

resources/lib/api.py
```diff
@@ -133,7 +133,7 @@ def handle_kodi_lookup_of_current_episode(self, tvshowid, current_episode_id):
                     position += 1
 
                 # now return the episode
-                self.log("Find current episode found episode in position: %s" % str(position), 2)
+                self.log("Find current episode found episode in position: %s" % position, 2)
                 try:
```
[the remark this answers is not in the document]
dagwieers: Not really, the %s casts it automatically to a string. So in this case I removed the useless cast.
But using '%d' might have worked as well.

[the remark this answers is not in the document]
dagwieers: In fact, using %d is cleaner, so I *will* fix it :-)

resources/lib/monitor.py
```diff
@@ -48,7 +49,7 @@ def run(self):
                 except Exception as exc:  # pylint: disable=broad-except
                     self.log("Exception in Playback Monitor Service: %s" % repr(exc))
 
-                    if 'not playing any media file' in str(exc):
+                    if 'not playing any media file' in from_unicode(exc):
```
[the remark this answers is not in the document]
dagwieers: True. Will fix.

resources/lib/monitor.py
dagwieers: ok.

resources/lib/playitem.py
dagwieers: This should stay `str()` in fact because we want to case an int to a string. I already fixed this.

dagwieers:
I think this is ready to be tested IRL (on Kodi 18 and Kodi 19, and Python 3).

Before we plan to do a release, I would make a list of add-on authors that integrate with Up Next, and ask them to test thoroughly as well.

mediaminister:
This is not ready to be merged, it doesn't work on Kodi 19 Python 3

im85288:
Oops @dagwieers could you send another PR with this reverted?

dagwieers:
I would simply fix the Python 3 issues we find. I was just going to test on a recent Kodi build on Windows.

mediaminister:
The play notification from service.upnext is not sent on Kodi 19 Python 3, I can take a look at it tomorrow.

dagwieers:
@mediaminister And I assume it worked before fine? (I never tested UpNext on Kodi 19 or Python 3 before)

dagwieers:
Ok, I find the error, but because of how this is logged, it is hard to know exactly where this issue happened. Personally I think we need to log the complete traceback in the future. I will add a fix for that too.

dagwieers:
It appears that the official v1.0.3 is also affected by this on Python 3, but it fails even sooner.
And Windows is always getting in the way, I hate it.

mediaminister:
@dagwieers I think everything is fixed now in master branch.

dagwieers:
Thanks, my Sunday didn't plan out as I expected 🙄

PR #62, opened 2019-10-19
Simplify summary/description and settings.xml

dagwieers:
I noticed both the summary and description where showing the same
information. Now the summary is a lot shorter, and the details are in
the description.

For the settings, I removed the default enable/visible defaults.


dagwieers:
I would even add a list of (maybe high-profile) add-ons that make use of this service add-on. Or ypou could even opt to show all existing add-ons that have this integration.

PR #68, opened 2019-10-20
Assorted fixes

dagwieers:
This PR includes:
- Replacing JSONRPC class with jsonrpc()
- Replace executebuiltin() with jsonrpc()
- Remove inspect.stack (would always return 'log')
- Make main service loop more efficient
- Get rid of broad exceptions
- Fix the wrong 'not playing any media file' string in exception
- Some cosmetic consistency
- Added class and method documentation (a lot is still needed)


dagwieers:
I often forget to create a PR as a Draft pull request, and GitHub doesn't allow to fix that retroactively. But now I did :-)

im85288:
Cool so I will not merge this one just yet :)

dagwieers:
This is ready for me, tested on Windows 10 with Kodi v19 and Python 3 as well.

dagwieers:
It's a lot more comprehensive than I originally wanted, but this is easier than 5 smaller PRs that conflict, both in managing as well as testing.

im85288:
Awesome thanks, the code is way more understandable

dagwieers:
I still don't understand why my codecov.yml is not being picked up.

I tried as `.github/codecov.yml` as well as `.codecov.yml` and both should work fine (and we do this elsewhere). But in this project is reports there is no codecov.yml, even though the syntax is tested and reported to be valid. (Sure, we use the same file elsewhere too)

```
$ cat .codecov.yml | curl --data-binary @- https://codecov.io/validate
Valid!

{
  "comment": false, 
  "ignore": [
    "^test/.*"
  ], 
  "coverage": {
    "status": {
      "project": {
        "default": {
          "threshold": 6.0, 
          "target": 75.0
        }
      }, 
      "patch": false
    }, 
    "range": [
      50.0, 
      100.0
    ], 
    "round": "nearest"
  }
}
```

im85288:
hmm no idea at all! Not used it before or indeed not seen it before

dagwieers:
So Codecov.io eventually panned out. We have a coverage of 34.7% just by running *service.py*.
What we lack is a bunch of unit tests, and the proper changes to the stubs so we can have integration tests that fake playing files under different circumstances...

PR #71, opened 2019-10-23
Move all media to resources/media/

dagwieers:
It is common to have all add-on media moved to resources/media.


dagwieers:
I did notice that the size of the screenshots (mostly) makes the ZIP package quite large (4.5MB in size). The content from *resources/media/* is 4.8MB in total.

Also, I don't know why these two images are in our source tree, as they do not seem to be used:
- upnext-developer.jpg
- upnext-settings.jpg

im85288:
Yes indeed they are there only for the wiki

PR #76, opened 2019-10-30
Support Back button action

### LINKED ISSUE #73: Make the Back button dismiss the pop-up without stopping
I think it makes sense to have the Back button dismiss the pop-up without stopping the currently playing file.

dagwieers:
The user should be able to use the back button to dismiss Up Next.
This behaves as if nothing had happened, except that it is clear that
the user is still watching. It also results in Up Next not playing the
next episode.

This fixes #73

### DIFF
--- a/resources/lib/stillwatching.py
+++ b/resources/lib/stillwatching.py
@@ -7,6 +7,7 @@ import xbmcgui
 from .statichelper import from_unicode
 
 ACTION_PLAYER_STOP = 13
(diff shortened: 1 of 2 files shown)


dagwieers:
@im85288 I noticed you merged #78, and not this one ;-) Is there something wrong with the approach in this PR?

im85288:
Nope all good! Just missed that this one was ready :)

PR #77, opened 2019-10-30
Reset resumetime before playing

### LINKED ISSUE #64: When next episode was already partly watched, a selection comes up
So I noticed that if the next episode was already partly watched, Next Up causes a selection window to show in the background with the option to **Resume from 00:00:50** or **Play from beginning**.

But it appears grayed out in the background while the screen is waiting to start playing. Very weird.

dagwieers:
This avoids the file being played to open a selection window to resume
from a specific position or start from the beginning.

This fixes #64 

Please test !


dagwieers:
Even though I have seen this a few times now, I can't seem to reproduce this. Neither for local media, as for add-ons. So don't merge this PR unless I can reproduce this again.

dagwieers:
This is now confirmed to be working. With the new settings we can actually use a large notification time (i.e. 60 secs) for very short episode (e.g. 7 mins) to easily test this.

The default cutoff time for Marking as Watched is 8% which would be 32 secs for a 7 min episode. So if you would click on "Watch Now" the current episode would not be marked as watched, but be marked as unfinished.

This is now fixed, and on top of that, batched JSONRPC queries work as well and are very efficient.

resources/lib/api.py
```diff
@@ -32,8 +32,19 @@ def addon_data_received(self, data, encoding='base64'):
...
+        # jsonrpc(dict(method='VideoLibrary.SetEpisodeDetails', params=dict(episodeid=episode.get('episodeid'), resume=dict(position=0))))
+
+        # Play the next episode
+        jsonrpc(dict(method='Player.Open', params=dict(item=dict(episodeid=episode.get('episodeid')))))
+
+        # If we do this too quickly, it won't be effective :-(
+        sleep(200)
+
+        # Mark the current episode as watched and reset position
+        jsonrpc(dict(method='VideoLibrary.SetEpisodeDetails', params=dict(episodeid=current_episodeid, playcount=1, resume=dict(position=0))))
```
dagwieers: We probably want to increase the existing playcount, rather than set it to 1 to indicate watched.

im85288: Not to sure, from memory I think Kodi only cares about it being either 0 or 1 and I think some addons possibly check for it actually being 1 to mark as watched.

dagwieers: Correct if you only consider the Watched flag. But I just recently made a pull request to add this information to the Video info pane so it is exposed to the user in Kodi Matrix :roll_eyes: 

So we might as well do the right thing here. Especially since we already have the playcount IIRC.

im85288: Cool so my memory was correct! Yeah let’s go for it then 😉

notoco:
I've just noticed that I've been having this problem, too - I'm looking forward to turning on this PR

dagwieers:
@notoco This is just for local media, we are waiting for changes in Kodi (see xbmc/xbmc#17202) to be accepted. That would allow us to do the same for VOD content (plugin:// URLs). Which would be a neat feature for Up Next to have!

PR #78, opened 2019-10-31
 Close button label

dagwieers:
This PR includes:
- If configured, relabel the Close button as Stop
- Use spaces not tabs for nextup XML, no other changes, just consistency


dagwieers:
I assume that for the StillWatching dialog it was a deliberate decision not to Stop when canceling (even when configured). So I didn't change StillWatching dialog.

If this is a bug, let me know and I will fix that.

PR #82, opened 2019-11-05
Fix IndexError: list index out of range

dagwieers:
This fixes the following issues:
```
File "/storage/.kodi/addons/service.upnext/resources/lib/api.py", line 201, in find_next_episode
  while result.get('result').get('episodes')[position].get('file') == current_file:
IndexError: list index out of range
```
And cleans up some of the logic.

### DIFF
--- a/resources/lib/api.py
+++ b/resources/lib/api.py
@@ -106,14 +106,10 @@ class Api:
             sort=dict(method='episode'),
         ))
 
-        if not result:
+        if not result.get('result'):
             return None
 
         self.log('Got details of next up episode %s' % result, 2)
-
-        if not result.get('result', {}).get('episodes'):
-            return None
-
         xbmc.sleep(100)
 
         # Find the next unwatched and the newest added episodes
@@ -128,43 +124,29 @@ class Api:
             sort=dict(method='episode'),
         ))
 
+        if not result.get('result'):
+            return None
+
         self.log('Find current episode called', 2)
-        if not result:
-            return None
-
-        # Find the next unwatched and the newest added episodes
-        if not result.get('result', {}).get('episodes'):
-            return None
-
         xbmc.sleep(100)
 
-        position = 0
-        for episode in result.get('result').get('episodes'):
+        # Find the next unwatched and the newest added episodes
+        episodes = result.get('result', {}).get('episodes', [])
+        for idx, episode in enumerate(episodes):
             # Find position of current episode
             if current_episode_id == episode.get('episodeid'):
-                # Found a match so get out of here
-                break
-            position += 1
+                self.log('Find current episode found episode in position: %d' % idx, 2)
+                return episode
 
-        # Now return the episode
-        self.log('Find current episode found episode in position: %d' % position, 2)
-        try:
-            episode = result.get('result').get('episodes')[position]
-        except (IndexError, KeyError) as exc:
-            # No next episode found
-            episode = None
-            self.log('error handle_kodi_lookup_of_current_episode %s' % repr(exc), 1)
-
-        return episode
+        # No next episode found
+        self.log('No next episode found', 1)
+        return None
 
     @staticmethod
     def showtitle_to_id(title):
         result = utils.jsonrpc(method='VideoLibrary.GetTVShows', id='libTvShows', params=dict(properties=['title']))
 
-        if not result.get('result', {}).get('tvshows'):
-            return '-1'
-
-        for tvshow in result.get('result').get('tvshows'):
+        for tvshow in result.get('result', {}).get('tvshows', []):
             if tvshow.get('label') == title:
                 return tvshow.get('tvshowid')
         return '-1'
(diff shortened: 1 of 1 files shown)


dagwieers:
I think this new logic is not only easier to read, it is also more correct in how it progresses through the episodes list. First going for a match, then proceeding to the next suitable candidate in the list.

@mediaminister @im85288 Please test.

resources/lib/api.py
```diff
@@ -178,38 +160,30 @@ def get_episode_id(showid, show_season, show_episode):
...
-                # Found a match so get out of here
-                break
-            position += 1
-
-        # Check if it may be a multi-part episode
-        while result.get('result').get('episodes')[position].get('file') == current_file:
-            position += 1
-
-        # Skip already watched episodes?
-        while not include_watched and result.get('result').get('episodes')[position].get('playcount') > 1:
```
dagwieers: This looked to me as a bug as `playcount > 1` means it has been watched at least twice, not just once.

resources/lib/api.py
```diff
@@ -178,38 +160,30 @@ def get_episode_id(showid, show_season, show_episode):
...
-            position += 1
-
-        try:
-            episode = result.get('result').get('episodes')[position]
-        except (IndexError, KeyError) as exc:
```
dagwieers: The IndexError and KeyError at this stage is useless as we may already have accessed it.
This is what needed fixing.

resources/lib/api.py
im85288: I'm not sure maybe Kodi defaults to having the playcount as 1

Logically I agree with you but I'm sure we had it in there for a reason or have you confirmed the items in the kodi database that are not watched have a playcount of 0?

dagwieers: Playcount really is what it means.
https://forum.kodi.tv/showthread.php?tid=300923

im85288: Ok cool let's go for it...will merge these changes in now

dagwieers: (I just made a small improvement.)
But I would prefer if this gets a bit more testing.

im85288: No problem I'll hold off for now, soon I'll make a pre-release and upload it to the Emby repository where there are a fair few candidates for testing..prior to releasing to the kodi repo

dagwieers: I always felt that Kodi would benefit from a wider testing community so that add-on developers can run test-releases of add-ons one or two weeks before they are released to a wider community. It would have prevent the urllib3 debacle.

A single library update could cause a lot of havoc for users. (Up Next is not at risk though)

dagwieers:
@mediaminister I know I ask a lot from you these days, but in case you have some spare time, please try this too.

resources/lib/api.py
dagwieers: BTW I still would like to see better integration tests to test fake playback and Up Next behaviour. But it means we have to implement a lot more stub functionality in the Player().

im85288: Maybe we can have a "TestPlayer" that has all the stubbed functionality and inject that when running the tests...similar ish to the Developer mode. I'm really a Java guy so not sure how/if that dependency can be injected at runtime etc

dagwieers: Well, I would like to test them in test/ without requiring Kodi. So we don't even need TestPlayer, just a stub implementation of the test/xbmc.Player is sufficient. I think the difficulty here will be to have this Player run in a thread, then instruct it from the integration tests to start playing, then seek, communicate, and then find out whether Up Next has performed what we expect (based on the configuration set). And for each test we need to reset the Player, and ultimately query its state.

I was a lot more optimistic before I looked at how it worked :-) But I still plan to do this, it's just not happening before making the improvements :-/

dagwieers: Hmm, I wanted to test this specific change, and I noticed that this doesn't work. Thinking about this, it is the add-on that proposes the next episode to be played. So this can only work for local files. Am I correct?

In this case I need to take this into account when redesigning the settings, because this is not clear for the user at this time.

I guess we need to make a distinction in the settings between options that affects local playback only , affects some add-ons and affects all add-ons.

dagwieers: It would also be nice if the add-on had an icon to access the settings from the add-ons menu.

dagwieers: I can confirm that the new implementation works for local playback. I then tried an older implementation, and it failed to work if the playcount was 1. What is strange here is that if the playcount was 1, it did not even propose to play anything. Whereas if the playcount was 2 it would jump over that episode.

So I think we indeed fixed a bug.

dagwieers: So maybe we should merge this as this now has become a bugfix PR :-)

mediaminister:
I never experienced this bug. I tested this branch with VRT NU add-on and Netflix add-on and Up Next works just like before.

dagwieers:
IIRC this issue can only really happen on local files.

dagwieers:
I think this is ready to be merged.

PR #90, opened 2019-11-23
convert bytes to str before using json.loads for python < 3.6 compat

anxdpanic:
Found this on my device with OSMC installed,  looks like OSMC is using Python 3.5 (at least currently)

- fixes utils.decode_data for `Python 3.0 -> 3.5`
- json.loads didn't support bytes or bytearray until Python 3.6

### DIFF
--- a/resources/lib/utils.py
+++ b/resources/lib/utils.py
@@ -78,7 +78,11 @@ def decode_data(data):
     ''' Decode data coming from a notification event '''
     data = json.loads(data)
     if data:
-        return to_unicode(json.loads(unhexlify(data[0])))
+        json_data = unhexlify(data[0])
+        # NOTE: With Python 3.5 and older json.loads does not support bytes or bytearray
+        if isinstance(json_data, bytes):
+            json_data = json_data.decode('utf-8')
+        return to_unicode(json.loads(json_data))
     return None
 
 


im85288:
@dagwieers Can you review this please

dagwieers:
I am surprised that OSMC would be shipping Python 3.5 (sep 2015) when Kodi only now supports Python 3. That makes no sense. I would be expecting Python 3.7 (jun 2018) or even Python 3.8 (oct 2019).

Python 3.5 support ends in 10 months.
https://devguide.python.org/#status-of-python-branches

dagwieers:
I cannot find any specific information to what Python versions are expected to be supported in Kodi v19. https://forum.kodi.tv/showthread.php?tid=339489

I would really prefer we don't need to support Python 3.5 at this point.

@Rechi @ronie @romanvm Please advise.

anxdpanic:
https://github.com/xbmc/xbmc/blob/2cd5ab12ae9a13231590b51a3cebe57e5c12de78/cmake/modules/FindPython.cmake#L8-L14

Rechi:
In general Kodi supports Python 3.5 to 3.7.
On depends platforms (Android, macOS, iOS) and Windows currently Python 3.7(.x) is used.
For Linux the distribution will decide which Python version is used for Kodi. That's also the reason why OSMC uses Python 3.5, because it's based on Debian Stretch, https://packages.debian.org/stretch/python3.

dagwieers:
@Rechi Thanks.

However I was expecting OSMC to still be using Python 2 as Python 3 support was only recently merged. What will happen when Python 3.5 is out of support, will Kodi drop support? And how does Debian Stretch cope with this? cc @samnazarko

resources/lib/utils.py
```diff
@@ -78,7 +78,10 @@ def decode_data(data):
     ''' Decode data coming from a notification event '''
     data = json.loads(data)
     if data:
-        return to_unicode(json.loads(unhexlify(data[0])))
+        json_data = unhexlify(data[0])
+        if isinstance(json_data, bytes):
```
dagwieers: Can you add a small comment that this is required for Python 3.5 and older ?
At least we know why this is needed and can remove it when it no longer is needed.

anxdpanic:
@dagwieers This is in the OSMC nightlies, sorry I didn't clarify

dagwieers:
@anxdpanic No worries, I am trying to wrap my head around this new Python 3.5 situation that may also affect anything integrating with Up Next (using hexlify as a result).

resources/lib/utils.py
anxdpanic: Comment is added 👍

anxdpanic:
I have it working on now on 3.5+. 
I've never mentioned the add-on anywhere though so there are very few users.

https://github.com/anxdpanic/plugin.video.composite_for_plex/blob/master/resources/lib/composite_addon/monitor.py#L30
https://github.com/anxdpanic/plugin.video.composite_for_plex/blob/master/resources/lib/composite_addon/common.py#L282

im85288:
Thanks guys merging this in now

samnazarko:
@dagwieers We'll start moving to Debian Buster before the end of the year, so we won't be shipping Kodi builds with Python 3.5 for long.

PR #103, opened 2019-12-01
Add a program entry in the addon section

dagwieers:
This adds an entry in the add-on section for Up Next.


im85288:
Cool, I’ve not been feeling great but will get round to doing a release this week sometime

dagwieers:
Not sure if we want to do this, we do this for InputStream Helper, but the use-case is different.

im85288:
If it makes finding the addon easier for end users that’s a positive

dagwieers:
Please release v1.0.6, keep these improvements for the next version.

dagwieers:
> If it makes finding the addon easier for end users that’s a positive

I agree.

im85288:
Will do thanks

PR #105, opened 2019-12-01
Add new icon with smaller logo

### LINKED ISSUE #104: White background on icon is really strong
I noticed that in the past there used to be a black background version of the icon, which I think is a better fit than the white box. Also making the icon itself somewhat smaller will be visually more appealing than having it as large as possible.

This is now also more important since we have an add-on icon in the main Program add-ons listing.

dagwieers:
This gets rid of the white background which is quite bright when using the default skin.

This fixes #104


anxdpanic:
Icons with transparency will result in a kodi-addon-checker error
```ERROR: Icon.png should be solid. It has transparency.```

dagwieers:
Good point, let me make it black then.

PR #107, opened 2019-12-02
Fix issue with 24hour format detection

### LINKED ISSUE #106: UnicodeDecodeError for Asian Language such as CJK
OS: MacOS 10.12.6
Kodi : Version 18.4. (20190831-3ade758ceb)
UpNext : 1.0.6
```
2019-12-02 21:55:09.484 T:123145514586112   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                            Error Type: <type 'exceptions.UnicodeDecodeError'>
                                            Error Contents: 'ascii' codec can't decode byte 0xec in position 0: ordinal not in range(128)
                                              File "/Users/allen/Library/Application Support/Kodi/addons/service.upnext/service.py", line 8, in <module>
                                                Monitor().run()
                                              File "/Users/allen/Library/Application Support/Kodi/addons/service.upnext/resources/lib/monitor.py", line 68, in run
                                                self.playback_manager.launch_up_next()
                                              File "/Users/allen/Library/Application Support/Kodi/addons/service.upnext/resources/lib/playbackmanager.py", line 40, in launch_up_next
                                                self.clock_twelve = 'm' in xbmc.getInfoLabel('System.Time').lower()
                                            UnicodeDecodeError: 'ascii' codec can't decode byte 0xec in position 0: ordinal not in range(128)
```
Please advise

dagwieers:
This is just a workaround because I don't think the implementation is
the best way to see if the clock is using a 24h format or not.

This fixes #106

### DIFF
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -10,7 +10,6 @@ from .api import Api
 from .player import Player
 from .playitem import PlayItem
 from .state import State
-from .utils import get_global_setting
 
 
 class PlaybackManager:  # pylint: disable=invalid-name
@@ -22,7 +21,6 @@ class PlaybackManager:  # pylint: disable=invalid-name
         self.play_item = PlayItem()
         self.state = State()
         self.player = Player()
-        self.clock_twelve = None
 
     def log(self, msg, level=2):
         utils.log(msg, name=self.__class__.__name__, level=level)
@@ -38,7 +36,6 @@ class PlaybackManager:  # pylint: disable=invalid-name
                 self.log('Error: no episode could be found to play next...exiting', 1)
                 return
         self.log('episode details %s' % episode, 2)
-        self.clock_twelve = bool(not get_global_setting('locale.use24hourclock'))
         self.launch_popup(episode, playlist_item)
         self.api.reset_addon_data()
 
@@ -109,7 +106,8 @@ class PlaybackManager:  # pylint: disable=invalid-name
             if episode_runtime:
                 end_time = total_time - play_time + episode.get('runtime')
                 end_time = datetime.now() + timedelta(seconds=end_time)
-                end_time = end_time.strftime('%I:%M %p' if self.clock_twelve else '%H:%M').lstrip('0')  # Remove leading zero on all platforms
+                time_format = xbmc.getRegion('time').replace(':%S', '')  # Strip off seconds
+                end_time = end_time.strftime(time_format).lstrip('0')  # Remove leading zero on all platforms
             else:
                 end_time = None
             if not self.state.pause:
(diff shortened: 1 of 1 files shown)


dagwieers:
@im85288 Please don't release without this final fix included :disappointed:

im85288:
Just let me know when it’s ready to merge and I’ll include it, we will need to make a 1.0.07 release but I’ll take care of that

resources/lib/playbackmanager.py
```diff
@@ -37,7 +38,7 @@ def launch_up_next(self):
                 self.log('Error: no episode could be found to play next...exiting', 1)
                 return
         self.log('episode details %s' % episode, 2)
-        self.clock_twelve = 'm' in xbmc.getInfoLabel('System.Time').lower()
+        self.clock_twelve = bool(not get_global_setting('locale.use24hourclock', True))
```
Thunderbird2086: I'm not sure about the second parameter for `get_global_setting()` since I see below in [utils.py](https://github.com/im85288/service.upnext/blob/16625400ec961add172428d6fabde91373bc064d/resources/lib/utils.py#L169)
```
169 def get_global_setting(setting):
```

dagwieers: Correct, our implementation of get_global_setting doesn't accept a default value. No worries, the new implementation gets rid of everything :-)

dagwieers:
@im85288 @mediaminister Please test.

Also voice your opinion on #104, so maybe, just maybe, we can include it in v1.0.7 too :-)

mediaminister:
It doesn't crash and it looks okay.

dagwieers:
@mediaminister I wasn't sure if the `getRegion()`-call is better placed outside of the loop.

In theory the user could change the locale settings during playback, but I am unsure if this would affect the `getRegion()` call. Also, I haven't timed the impact of a `getRegion()` call, but would expect this to be negligible.

Thunderbird2086:
@dagwieers I already tested the latest PR and I don't see any issue 👍

mediaminister:
The loop is only executing during the "Up Next" notification, the last 30 seconds.
I didn't see any issue with "Up Next" while changing locale trough JSON-RPC
```
curl -X POST -H "content-type:application/json" http://localhost:8080/jsonrpc -d '{"jsonrpc": "2.0", "method": "Settings.SetSettingValue", "params": {"setting": "locale.country", "value": "Central Europe"}, "id": 1}'
```

```
curl -X POST -H "content-t:8080/jsonrpc -d '{"jsonrpc": "2.0", "method": "Settings.SetSettingValue", "params": {"setting": "locale.country", "value": "USA (12h)"}, "id": 1}'
```
The `end_time` format is changed realtime, but I think the window property `endtime` is not used in the standard skin.

dagwieers:
So I guess this is ready to be merged.

PR #109, opened 2019-12-02
Use formatted sentences in translations

### LINKED ISSUE #108: Suggestion:  single string for "Up Next in" for notification
As ["Up Next in"](https://github.com/im85288/service.upnext/blob/16625400ec961add172428d6fabde91373bc064d/resources/skins/default/1080i/script-upnext-upnext.xml#L100) is declared in msgctxt `#30008` and `#30036` separately, it's hard to add translation for Japanese. 

For Japanese, the translation becomes as below. 
```
次のエピソードの再生までsss秒

   where sss is seconds
```
As there's no space in Japanese,  the separated `in`(#30036) cannot be translated properly.

dagwieers:
Since some languages do not translate like most western languages
(composed of substrings) it is better to translate sentences, rather
than substrings.

This is required for Korean and/or Japanese translations.

But it is generally a better practice anyway.

This fixes #108

### DIFF
--- a/resources/language/resource.language.de_de/strings.po
+++ b/resources/language/resource.language.de_de/strings.po
@@ -108,16 +108,16 @@ msgid "Close"
 msgstr "Schließen"
 
 msgctxt "#30035"
-msgid "seconds"
-msgstr "Sekunden"
+msgid "Continue watching in [COLOR FFFF4081]$INFO[Player.TimeRemaining(ss),,][/COLOR] seconds"
+msgstr "Wiedergabe fortsetzen in [COLOR FFFF4081]$INFO[Player.TimeRemaining(ss),,][/COLOR] Sekunden"
 
 msgctxt "#30036"
-msgid "in"
-msgstr "in"
+msgid "Up next in [COLOR FFFF4081]$INFO[Player.TimeRemaining(ss),,][/COLOR] seconds"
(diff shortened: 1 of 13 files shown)


dagwieers:
@Thunderbird2086 This should get you going again. I need to finish off the other languages before it can be merged. And we may want to delay merging until v1.0.7 is released. (v1.0.7 has important fixes that currently breaks various add-ons, so it has highest priority)

Your translation and these fixes, and the new settings, can go into v1.1.0

dagwieers:
@im85288 @mediaminister This one is ready to be reviewed. But probably best for a v1.1.0 release.

dagwieers:
Maybe we can start preparing v1.1.0 in the devel branch now that v1.0.7 is out?

Any future changes to the v1.0 releases could go in special temporary branch, if needed.
(Hopefully not)

mediaminister:
> This one is ready to be reviewed.

This works just fine, tested on Kodi 18 and Kodi 19.

im85288:
Yeah I guess using the dev branch would make more sense. I usually avoid as it just adds to the complexity but probably worthwhile here. Still not very well at the moment so maybe in a couple of days I’ll be able to take a look

dagwieers:
@im85288 If it would help, I don't mind doing triage and roadmap management (assigning issues/PRs to releases). Not sure if GitHub has a different role for this, I would expect so.

PR #111, opened 2019-12-03
Make icon size 512x512 as needed

dagwieers:
As requested by Kodi.


dagwieers:
I recreated it from the original to get the best quality.
So I didn't rescale it from 500x500 to 512x512.

PR #100, opened 2019-11-28
Implement new settings

dagwieers:
This implements reorganized settings:
- 3 sections: Interface, Behaviour, Expert
- Customize the notification time by episode length
- Various improvements to translation strings

The changes are in the second commit (adf6dd08bee69937c34948cdc5d2b95604d60c7f), the first commit is PR #114.

The new settings look like this:
screenshot031
screenshot036
screenshot037
screenshot038

PR #114, opened 2019-12-04
Restructure some code

dagwieers:
These are the non-impacting changes from PR #100.

This PR includes:
- Selectively import classes and functions from modules
- Rename 'settings()' into 'get_setting()'
- Change 'window()' into 'clear_property()', 'get_property()' and 'set_property()'
- Use Addon instance on demand so settings are updated instantly
- No longer require a restart/reboot for settings to be applied


dagwieers:
By moving this out of PR #100 it will make PR #100 a lot easier to understand.

PR #116, opened 2019-12-05
Remove shortPlay functionality

dagwieers:
This was already disabled from the settings after #100 was merged.
So I guess we can remove this functionality.

Users can now only disable the Up Next pop-up for video's smaller than
10mins by configuring a notification time of 0.


dagwieers:
This was not discussed in detail in PR #100 
We removed this functionality from the settings, not sure if we want to retain some of it?

im85288:
Happy to remove it now that the settings cover more use cases

PR #118, opened 2019-12-05
Replace Developer mode with testing a specific window

dagwieers:
Since Developer mode requires playing a file and this may confuse users
accidentally enabling developer mode, it may be easier if we simply show
the selected window (pop-up) from the settings directly.

This PR includes:
- Removing Developer mode in favor of testing popup windows
- Move script and service entry points into resources/lib/
- Add script.py

This fixes #86 and #120

Issue #117, opened 2019-12-05
Not advancing to next episode


jgslade:
At the end of an episode instead of going to the next one it is now staying on the last frame and the timer keeps going. 

I am running on libreelec 9.2 with Kodi 18.5. I have rebooted the system, powered down the system, and uninstalled up next both with and without deleting the settings.

im85288:
Please see: https://github.com/im85288/service.upnext/wiki/Report-a-Bug for how to report a bug with the info we need to try and resolve it

dagwieers:
@im85288 We can make this easier by adding an issue-template in GitHub with the instructions and the information we expect.

I will make a proposal.

notoco:
Sometimes I wonder if you're even sleeping...

jgslade:
https://paste.kodi.tv/abizirojaw.kodi

UpNext is version 1.0.7
Libreelec has the latest updates installed.

When the log was taken the show was at minute 24 of about 21.

I enabled debug on the UpNext addon settings, do I need to change any Kodi settings to get a more meaningful log?

im85288:
No the logs have all the info we need...however it seems to say everything worked correctly. Let’s see if @dagwieers has any ideas

dagwieers:
I don't see a good reason why it would have failed for local media. I explicitly tested this user scenario before the v1.0.7 release, and I just tested it again with the v1.0.7 release and it works fine for me. I don't know what else to try to debug this.

(Tested on LibreELEC v9.2 on Raspberry Pi)

dagwieers:
One thing you could maybe try is to enable debug logging in Kodi.
Maybe the issue is in Kodi itself?

And you could try and reproduce it using a previous version of Up Next (like v1.0.1).

jgslade:
The media is an NFS share.

Here is a link to the log after enabling logging in kodi:
http://ix.io/23DP

I just tried it with 1.0.1 and it worked. Here is the log from it working:
https://paste.kodi.tv/yudacewifa.kodi

dagwieers:
> The media is an NFS share.

My media as well.

dagwieers:
I am sorry, I can't tell anything from the v1.0.7 logs. It appears to play it correctly, and the Kodi logging is simply too verbose and doesn't give me any obvious clues.

The v1.0.1 logs don't log anything coming from the add-on, so not sure what is going on there.

jgslade:
> My media as well.

Drat

[...]

Issue #86, opened 2019-11-17
Clicking Play Next or Close does nothing


stathis95194:
I cannot manage to make this work. All settings to Up Next and Seren or Venom are enabled and identical. Popup comes up as expected on the time expected, but even if click Play Next, nothing happens. Stream finishes and continues to the next one only when time runs out.

Log isn't very helpful as it shows that everything is fine even though it isn't. Am i missing something?

A thing i noticed is that at the beginning of each episode, Up Next shows the popup for a completely unrelated show (Game of Thrones s2e4) which i need to "Close" 3 times to disappear. I found nothing on log regarding this though

`2019-11-17 19:50:08.468 T:14427  NOTICE: Opening stream: 0 source: 1024
2019-11-17 19:50:41.590 T:12617  NOTICE: UP NEXT Monitor -> log : Calling autoplayback totaltime - playtime is 49.5129394531
2019-11-17 19:50:41.603 T:12617  NOTICE: UP NEXT Api -> log : handle_addon_lookup_of_next_episode episode returning data {"rating": 8.0, "tvshowid": "tt4524056", "art": {"tvshow.fanart": "https://www.thetvdb.com/banners/fanart/original/294741-8.jpg", "thumb": "https://www.thetvdb.com/banners/episodes/294741/5202823.jpg", "tvshow.clearart": "", "tvshow.landscape": "https://www.thetvdb.com/banners/fanart/original/294741-8.jpg", "tvshow.poster": "https://www.thetvdb.com/banners/posters/294741-15.jpg", "tvshow.clearlogo": ""}, "plot": "Macarena's confrontation with self-appointed prison queen bee Zulema endangers her family.", "season": 1, "title": "Episode 3", "episodeid": 1848870, "firstaired": "2015-05-04", "episode": 3, "playcount": 0, "showtitle": "Vis a Vis"}
[...]

PR #119, opened 2019-12-05
Update Italian translation

EffeF:


dagwieers:
I think it is better to wait until we are ready to release, because we are still adding/modifying functionality that influences translations.

im85288:
Good point not sure I can un merge but for future ones will keep them open until things are settled

dagwieers:
I think it is best we ask translators to update translations.
And do this a few days before we expect a release.

I was planning to do this and call out to a lot of new translators as well, but I waited until #100 because of the impact. And now with #118 there's again more impact.

im85288:
Fair enough, there’s also now been two reports of the Popup not showing so something we need to get to the bottom off

dagwieers:
Two ? I only know of one: #117 
Maybe because I think #86 is related to enabling a broken Developer mode.

im85288:
Someone posted in issue 18 (sorry cannot link) which is already closed so I asked them to create a new issue which they have not. Not much info though

dagwieers:
BTW You can simply type: &lt;hash&gt; &lt;issue number&gt; and it automatically creates a link. Like this: #18 

The Venom/Seren issue (#91) was reported even before we released v1.0.7. And they tested combinations of old/new Addon Signals and older/newer UpNext. So something else is going on there and I suspect a logic issue in the code. When I reviewed Venom's code I saw some practices that could potentially lead to issues unrelated to UpNext.

Issue #75, opened 2019-10-30
Update resumetime when Watch Now pressed


dagwieers:
So I have the following problem that when an episode is almost finished and I press "Watch Now" to go to the next episode, the episode is not "finished watching" in Kodi because it was not played until the end. (Or the cut-off time to consider it watched is not in sync with the Up Next interval).

I guess what is needed is a way to update this `ignorepercentatend` from the *advancedsettings.xml* in Kodi. This would require improvements to Kodi to either read and change this value, or make it part of the Kodi settings altogether.
https://kodi.wiki/view/HOW-TO:Modify_automatic_watch_and_resume_points

In the meantime we should document in the Wiki that the user should keep this in sync with the Up Next settings.

dagwieers:
I know why this happens for me so often. It seems for our add-on **VRT NU** we return a `notification_time` of 60 secs, which overrides the user-configured notification time of 30 secs.

This was confusing me before as well.

@mediaminister Why are we forcing a notification time of 60 seconds and not leave this up to the user? I think 60 secs is a long time for typical TV show episodes and it causes an episode to not be completely watched when pressed :-/ (And the kids apparently press **Watch Now** when Up Next comes up too ;-))

dagwieers:
It seems `Files.SetDetails()` (see #77) does work so we could fix this from within Up Next.

mediaminister:
I forced 60 seconds because the credits of most fiction series have a duration that is much longer than 30 seconds. As I didn't succeed to detect the timecode of starting credits, it's indeed better to let the user decide which notification is used.

dagwieers:
@mediaminister My PR has it now forced to 30 seconds, do you think we ought to make this configurable in the VRT NU add-on, or should we use the Up Next configured value ?

I can see the benefit of having this configurable for VRT NU specifically, as typically TV content does not have very long credits.

mediaminister:
No problem to make a config option in VRT NU add-on.

dagwieers:
I have to reconsider my opinion here (again). It is better to show the dialog earlier, than too late. When I was watching the first Punk episode (1h30) the end credits start 50 seconds before the end. Which makes tagging an episode as Watched even more important. And I think Kodi needs to be redesigned in that aspect. (advancedsettings globally is not a good approach)

To be honest, I think a better approach is to make the dialog pop up at different intervals depending on what we're playing and the length of the video.
- Videos less than 40mins: 15 secs (typically short shows, talk show, kids shows)
[...]

PR #133, opened 2019-12-10
Implement play_url support

### LINKED ISSUE #132: Support for a plugin URL as play info
I think we can simplify how the add-on integrates with UpNext.

Currently it uses notifications,
- one coming from the add-on for informing UpNext about a potentially upcoming episode for the currently playing video, and
- one from UpNext to ask the add-on to play the next episode, when that time has come

For the second notification it requires the add-on to have a service thread that waits for this notification and acts on it. This is complexity that should not be needed. In fact, requiring add-ons to have their own service is (in the larger picture) unwanted.

We could also support a mechanism that lets UpNext initiate the next episode itself. Let's call this `play_url` and it would contain e.g. `plugin://plugin.video.vrt.nu/play/upnext/vid-57e751d2-e744-4b33-ade6-b33ea3e1bba5` so that the add-on does not need any service running.

This would make integrating with UpNext a lot easier for most add-ons as they only need to perform a single notification.

dagwieers:
This adds support for playing the next episode directly from UpNext.

This fixes #132

### DIFF
--- a/resources/lib/api.py
+++ b/resources/lib/api.py
@@ -60,8 +60,12 @@ class Api:
         return item
 
     def play_addon_item(self):
-        self.log('sending %(encoding)s data to addon to play: %(play_info)s' % dict(encoding=self.encoding, **self.data), 2)
-        event(message=self.data.get('id'), data=self.data.get('play_info'), sender='upnextprovider', encoding=self.encoding)
+        if self.data.get('play_url'):
+            self.log('Playing the next episode directly: %(play_url)s' % self.data, 2)
+            jsonrpc(method='Player.Open', params=dict(item=dict(file=self.data.get('play_url'))))
+        else:
+            self.log('Sending %(encoding)s data to add-on to play: %(play_info)s' % dict(encoding=self.encoding, **self.data), 2)
+            event(message=self.data.get('id'), data=self.data.get('play_info'), sender='upnextprovider', encoding=self.encoding)
 
     def handle_addon_lookup_of_next_episode(self):
         if not self.data:
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -73,12 +73,14 @@ class PlaybackManager:
         self.log('playing media episode', 2)
         # Signal to trakt previous episode watched
         event(message='NEXTUPWATCHEDSIGNAL', data=dict(episodeid=self.state.current_episode_id), encoding='base64')
-        # Play media
         if playlist_item:
+            # Play playlist media
             self.player.seekTime(self.player.getTotalTime())
         elif self.api.has_addon_data():
+            # Play add-on media
             self.api.play_addon_item()
         else:
+            # Play local media
             self.api.play_kodi_item(episode)
 
     def show_popup_and_wait(self, episode, next_up_page, still_watching_page):
--- a/resources/lib/utils.py
+++ b/resources/lib/utils.py
@@ -13,7 +13,7 @@ ADDON = Addon()
 
 
 def get_addon_info(key):
-    ''' Return addon information '''
+    ''' Return add-on information '''
     return to_unicode(ADDON.getAddonInfo(key))
 
 


im85288:
@dagwieers  Cool looks straight forward enough to me, can you update the wiki with this information?

dagwieers:
Let's first test this and see what others think :-)

im85288:
No worries, It's additional functionality so I cannot think of any reason why anyone would have objections :)

dagwieers:
Right, but the implementation could have different options:
- Do we need to support additional data being added to the URL?
- Is `Player.Open()` really the best option, or do we use `RunPlugin()` instead?
- Or do we add the item to the playlist, jump to the end and play the next item on the playlist?
  - This appears to give the best user-experience we determined
  - This also ensures the video is properly tagged as being finished (one issue we had)

So I do think we have to ask other UpNext stakeholders to get the best implementation.

dagwieers:
So who are all the UpNext stakeholders? Feedback welcome!
cc @mediaminister @Castagnait @notoco @angelblue05 @123Venom @anxdpanic

CastagnaIT:
i have do a try seem to work good
so implementing this i found a problem in nf, as soon as i solve i will do a full test

dagwieers:
So in our add-on we have this implemented as a work-around to switching back to the virtual directory listing and back into the next episode (when it automatically moves to the next episode at the end of the current episode).

Since going back to the original virtual directory listing might be expensive, but also because it doesn't look nice, we opted to use the playlist for a perfect transition from current to next episode.

And this is something we could also add to Up Next, either by default or optionally:
```python
        # To have a perfect transition, add to the playlist if automatically continuing
        if self.isPlaying() and self.getTotalTime() - self.getTime() < 1:
            log(3, '[PlayerInfo] %d Add %s to Kodi Playlist' % (self.thread_id, url))
            PlayList(1).add(url)
        else:
            log(3, '[PlayerInfo] %d Add %s to Kodi Player' % (self.thread_id, url))
            self.play(url)
```

mediaminister:
I think this is a nice improvement as it simplifies add-on integration a lot!

> Or do we add the item to the playlist, jump to the end and play the next item on the playlist?

I'm in favor of using `xbmc.PlayList` so Kodi never returns to the virtual directory listing while using "Up Next". I'm convinced this is the most failsafe approach.

notoco:
@dagwieers - nie używam wtyczki UpNext w innych dodatkach - wszystko opieram o bibliotekę (wszystkie dodatki z których korzystam z niej korzystają), więc nie wiem czy to mi pomoże. Natomiast Netflix z poprawką + Upnext z play_url działają niesamowicie szybko...

CastagnaIT:
to netflix addon this is a breacking change, so please let me know, when after this is merged, the new version is released
so i merge the change to nf thanks!

im85288:
> to netflix addon this is a breacking change, so please let me know, when after this is merged, the new version is released
> so i merge the change to nf thanks!

Will do, in theory the old method should still work so not expected to break any current addon integration...but I will let @dagwieers confirm this

CastagnaIT:
yes both are supported, but i want to avoid doing version checks or other, i prefer to make the change by deleting the other method

im85288:
Ok cool when merged I will shout you out on here so that are you notified

dagwieers:
I don't want to rush this. To me it would be ideal if Up Next could not just perform initiating the next episode (the best way possible) but also ensures the current episode is marked as watched. Unfortunately that is currently unsupported by Kodi. (see #75 for the details)

dagwieers:
Another option is to allow an add-on to notify UpNext by calling `RunScript,()` instead of using `NotifyAll()`, however I don't see any compelling reasons for doing this. We still need to base64-encode the payload and I don't expect any speed improvements (it is not a critical path where timing is of the utmost importance). The goal is not to get rid of `NotifyAll()` per se.

dagwieers:
@mediaminister I tried various combinations of these:
```python
jsonrpc(method='Playlist.Insert', params=dict(playlistid=1, position=1, item=[dict(file=self.data.get('play_url'))]))
jsonrpc(method='Player.Seek', params=dict(playerid=1, value=100.0))
jsonrpc(method='Player.GoTo', params=dict(playerid=1, to='next'))
jsonrpc(method='Player.Open', params=dict(item=dict(file=self.data.get('play_url'))))
```

The idea was to:
1. Add the video to the Playlist
2. Seek to the end (to mark episode as watched)

Unfortunately, this doesn't seem to work very well.

Only `Player.Open` works very well if you press "Watch Now", if you wait until the end you first go to the virtual directory before opening the new episode. This is not worse than before for most add-ons though. So just doing `Player.Open` is not a regression, it just doesn't work as well as we are used to in VRT NU currently. (On the other hand, most people tend to click on "Watch Now" or "Close")

CastagnaIT:
if instead of waiting for the end, do you run the command 1/2 seconds before the end?
is a bad solution, but could it work?

mediaminister:
Seeking to the end triggers `onPlayBackEnded` and causes Kodi to return to the virtual directory.
That must in any case be avoided to seamlessly play the next episode.

dagwieers:
@CastagnaIT Well, currently we loop every 100ms, and we wait until totaltime-position < 1, so in theory that would be on average 950ms before the end. But there's some more processing in the loop and after the loop so I don't have exact timings (see playbackmanager.py).

That said, I think we can do with cleaning up some of the logic as there's quite some complexity (also related to naming or lack of comments). Same for #135

CastagnaIT:
100ms! this is absurd LOL for what reason?
this make so many unnecessary calls to kodi api

it is not enough to interrogate every 500ms or max 1sec? to the human eye this is not visible!
in this way wasted a lot of cycles in low-power CPU...

dagwieers:
@CastagnaIT Well, let's not overexaggerate. It has been like this since ever (?) and nobody complained. Feel free to investigate what would be better values and report back with a PR and test results. Maybe @im85288 can fill you in with the details of past choices.

My concerns are more related to better testing and understandable code. And I try not to make too big changes without the tests to back things up. And Up Next is lacking proper integration tests for starting to rewrite important bits IMO.

CastagnaIT:
because you mentioned but usually i don't study the other addons if it's not required
now that I've been looking at the code i think I've figured out the reason for the 100ms

im85288:
Guys I can’t recall the exact piece of code we are talking about here. Possibly it’s set to 100ms to make sure the end of the episode does not come up before the timer stops so that we can invoke playback of the next episode whilst the playback manager is still alive.

What’s the actual problem here?

CastagnaIT:
you distracted me, i'm watching foxes, no worry about my comment

im85288:
Haha ok no worries, enjoy the foxes

mediaminister:
@dagwieers I think your current implementation is trying to duplicate existing functionality.
I saw that "Up Next" already supports playing the next `xbmc.PlayList()` item in `playitem.py`

I got this perfectly working by editing the `get_next` function:

```python
    def get_next(self):
        playlist = PlayList(PLAYLIST_VIDEO)
        position = playlist.getposition()
        
        # Support play_url
        play_url = self.api.data.get('play_url')
        if play_url:
            self.log('Adding add-on play_url to playlist: %(play_url)s' % self.api.data, 2)
            playlist.add(self.api.data.get('play_url'))

        # A playlist with only one element has no next item and PlayList().getposition() starts counting from zero
        if playlist.size() > 1 and position < (playlist.size() - 1):
            if play_url:
                return self.get_episode()
            return self.api.get_next_in_playlist(position)
        return False
```

dagwieers:
@mediaminister It does not appear to be working here. Not sure what is different, but if I press "Watch Now" the video stalls.

Also, this appears to add the next episode to the playlist at the very start of the playback, but if the user chooses "Close" the item is still added to the playlist (and the playlist was modified). Whereas we would expect it to be added only when the user chooses "Watch Now".

dagwieers:
There also appears to be a problem where the next episode when played has issues with scrolling (in VRT NU). It appears as if  more than one thread is interfering with the position and skipping forward or backward does work for the video, but the timing/timebar jumps back. Very weird.

dagwieers:
> No, the episode is added at notification time in launch_up_next

You are right.

> You can easily make the "Close" button functional with playlists

But that would destroy the original playlist? IIUIC the existing playlist support is to ensure that Up Next advances to the next item in the playlist (not necessarily the next episode). So we probably don't want to mess with the playlist if the playlist was not already empty.

I am worried that it increases complexity of we are mixing these two use-cases using the same logic. (And that is also why I had hoped using jsonrpc would make things somewhat easier without requiring a redesign of Up Next)

Maybe we should merge the simple case using `Player.Open()` and look into improving this when reorganising the codebase (which we can in all honestly only do if we have better integration tests enforcing existing behaviour).

mediaminister:
> But that would destroy the original playlist?

That's right, a better option is removing the item that was just added.

> So we probably don't want to mess with the playlist if the playlist was not already empty.

Maybe, we certainly need to investigate all current use cases of this Playlist feature. There must be a way to make this work while preserving existing use cases.

> Maybe we should merge the simple case using Player.Open()

No, this branch is not ready for merging. I already got a timing issue with this implementation where a next episode failed to play.

notoco:
So you are waiting to be connected to this :( and with Netflix it works so well :)

mediaminister:
It doesn't work reliable with VRT NU add-on, here is a full debug log where `plugin://plugin.video.vrt.nu/play/upnext/vid-9f408f8e-df0c-4568-8615-8d74059a3e50` failed to play:
[kodilog.txt](https://github.com/im85288/service.upnext/files/3963839/kodilog.txt)

im85288:
@dagwieers where are we with this one?

dagwieers:
@im85288 No need to rush things. It seems to work very well for Netflix, but not for other add-ons. I would rather find a common solution that works well in all cases.

dagwieers:
@mediaminister If we would be releasing Up Next v1.1.0 this week, could we release with the original implementation of this PR (which was pretty straightforward and worked well for Netflix). Or do we postpone this hoping to get something better in?

mediaminister:
This needs more testing. I'm not sure this works reliable with VRT NU: https://github.com/im85288/service.upnext/pull/133#issuecomment-565713295

dagwieers:
@mediaminister We would not be using it for VRT NU, until we find something that Works-For-Everything&trade;.

mediaminister:
Okay, this can be merged. I totally forgot this is using a new API call.

dagwieers:
@im85288 Ready to be merged !

im85288:
Whoops ignore the thumbs down! Big fingers on a small phone!

Issue #139, opened 2019-12-17
Timing issues preventing Up Next working reliably


thebertster:
I am continuing to have issues with Up Next working reliably on both my Kodi systems (one is a Raspberry Pi 3, the other a slightly faster Intel Celeron fanless PC both running LibreElec 9.2.0/Kodi Leia 18.5).

I thought I'd try to debug the behaviour with the current version of the service pulled directly from the repo (last commit being f18864b on December 12).

There seems to be a timing issue with api.play_kodi_item() being invoked at the same time as the previous episode's playback is in the process of stopping. This appears to cause the next episode not to play - on my systems, this happens almost every time (maybe about 10% of the time the next episode does play).

The debug logs when this happens are as below. Playback of the previous episode has finished as can be seen by the VideoPlayer::OnExit() event being logged. You can then see the debug message "playing media episode" logged and then the VideoPlayer::OpenFile call is made. Playback of the next episode does not start - the VideoPlayer just exits.. The next episode never ends up playing successfully - I guess this is a Kodi VideoPlayer "issue" in that sense, but realistically it is only with something like Up Next that the precise timing needed to trigger the behaviour would be seen.

```
2019-12-16 23:35:54.556 T:140524635789056  NOTICE: [service.upnext] UpNextMonitor -> Show notification as episode (of length 1269 secs) ends in 15 secs
2019-12-16 23:35:54.559 T:140524635789056  NOTICE: [service.upnext] Api -> Got active player {u'jsonrpc': u'2.0', u'id': 1, u'result': [{u'playerid': 1, u'type': u'video', u'playertype': u'internal'}]}
2019-12-16 23:35:54.561 T:140524635789056  NOTICE: [service.upnext] Api -> Getting details of now playing media
2019-12-16 23:35:54.611 T:140524635789056  NOTICE: [service.upnext] Api -> Got details of now playing media {...removed JSON!...}
2019-12-16 23:35:55.711 T:140524635789056  NOTICE: [service.upnext] Api -> Got details of next up episode {...removed JSON!...}
2019-12-16 23:35:55.823 T:140524635789056  NOTICE: [service.upnext] PlaybackManager -> episode details {...removed JSON!...}
2019-12-16 23:35:55.829 T:140524635789056  NOTICE: [service.upnext] PlaybackManager -> played in a row settings 3
2019-12-16 23:35:55.831 T:140524635789056  NOTICE: [service.upnext] PlaybackManager -> played in a row 1
[...]

PR #142, opened 2019-12-23
Fix various runtime exceptions

thebertster:
Runtime exception in playbackmanager.py if neither next_up nor still_watching pages get shown because of early abort (e.g. user skipped directly to the end of the video).

Runtime exception when displaying either page if the skin does not have the progress control (ID 3014). Was checking for None on return of getControl() but this actually raises a RuntimeError exception if the control does not exist.

### DIFF
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -150,6 +150,7 @@ class PlaybackManager:  # pylint: disable=invalid-name
             should_play_default = still_watching_page.is_still_watching()
             should_play_non_default = still_watching_page.is_still_watching()
         else:
+            # FIXME: This is a workaround until we handle this better (see comments in #142)
             return False, False
 
         if next_up_page.is_watch_now() or still_watching_page.is_still_watching():
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -149,6 +149,8 @@ class PlaybackManager:  # pylint: disable=invalid-name
             still_watching_page.close()
             should_play_default = still_watching_page.is_still_watching()
             should_play_non_default = still_watching_page.is_still_watching()
+        else:
+            return False, False
 
         if next_up_page.is_watch_now() or still_watching_page.is_still_watching():
             self.state.played_in_a_row = 1
--- a/resources/lib/stillwatching.py
+++ b/resources/lib/stillwatching.py
@@ -57,9 +57,11 @@ class StillWatching(WindowXMLDialog):
             self.setProperty('playcount', str(self.item.get('playcount', 0)))
 
     def prepare_progress_control(self):
-        self.progress_control = self.getControl(3014)
-        if self.progress_control is not None:
+        try:
+            self.progress_control = self.getControl(3014)
             self.progress_control.setPercent(self.current_progress_percent)  # pylint: disable=no-member
+        except RuntimeError:
+            pass
 
     def set_item(self, item):
         self.item = item
@@ -69,9 +71,11 @@ class StillWatching(WindowXMLDialog):
 
     def update_progress_control(self, remaining=None, endtime=None):
         self.current_progress_percent = self.current_progress_percent - self.progress_step_size
-        self.progress_control = self.getControl(3014)
-        if self.progress_control is not None:
+        try:
+            self.progress_control = self.getControl(3014)
             self.progress_control.setPercent(self.current_progress_percent)  # pylint: disable=no-member
+        except RuntimeError:
+            pass
         if remaining:
             self.setProperty('remaining', from_unicode('%02d' % remaining))
         if endtime:
--- a/resources/lib/stillwatching.py
+++ b/resources/lib/stillwatching.py
@@ -59,9 +59,10 @@ class StillWatching(WindowXMLDialog):
     def prepare_progress_control(self):
         try:
             self.progress_control = self.getControl(3014)
-            self.progress_control.setPercent(self.current_progress_percent)  # pylint: disable=no-member
-        except RuntimeError:
+        except RuntimeError: # Occurs when skin does not include progress control
             pass
+        else:
+            self.progress_control.setPercent(self.current_progress_percent)  # pylint: disable=no-member
 
     def set_item(self, item):
         self.item = item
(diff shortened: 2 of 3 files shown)


im85288:
Thanks for this I’ll let @dagwieers have a look and look to merge in the next few days.

resources/lib/stillwatching.py
```diff
@@ -57,9 +57,11 @@ def set_info(self):
             self.setProperty('playcount', str(self.item.get('playcount', 0)))
 
     def prepare_progress_control(self):
-        self.progress_control = self.getControl(3014)
-        if self.progress_control is not None:
+        try:
+            self.progress_control = self.getControl(3014)
             self.progress_control.setPercent(self.current_progress_percent)  # pylint: disable=no-member
+        except RuntimeError:
```
dagwieers: Only use a try-block for the statement you know will throw the exception, if you still need to perform tasks when there is no exception, please use an else-block.

It is also a good idea to document when this specific exception is thrown, so people after you know the exact case you are covering.

So:
```python
        try:
            self.progress_control = self.getControl(3014)
        except RuntimeError:  # Happens when the skin has no progress bar control
            pass
        else:
            self.progress_control.setPercent(self.current_progress_percent)  # pylint: disable=no-member	            
```

dagwieers: Please fix this for the other occurrences as well. Thanks !

resources/lib/playbackmanager.py
```diff
@@ -149,6 +149,8 @@ def extract_play_info(self, next_up_page, showing_next_up_page, showing_still_wa
             still_watching_page.close()
             should_play_default = still_watching_page.is_still_watching()
             should_play_non_default = still_watching_page.is_still_watching()
+        else:
+            return False, False
```
dagwieers: I fail to understand when this would happen. I would expect this is run for either next-up or still-watching pop-ups, there should not be another scenario?

thebertster:
It occurs if player is no longer running by the time show_popup_and_wait() is called in playbackmanager.py, in which case an exception is caught and show_popup_and_wait() returns without showing either of the pop-ups. This happens for example if the user skips forward past the end of the playing episode.

```
    def show_popup_and_wait(self, episode, next_up_page, still_watching_page):
        try:
            play_time = self.player.getTime()
            total_time = self.player.getTotalTime()
        except RuntimeError:
            self.log('exit early because player is no longer running', 2)
            return False, False
```

resources/lib/playbackmanager.py
thebertster: It occurs if player is no longer running by the time show_popup_and_wait() is called in playbackmanager.py, in which case an exception is caught and show_popup_and_wait() returns without showing either of the pop-ups. This happens for example if the user skips forward past the end of the playing episode.

```
    def show_popup_and_wait(self, episode, next_up_page, still_watching_page):
        try:
            play_time = self.player.getTime()
            total_time = self.player.getTotalTime()
        except RuntimeError:
            self.log('exit early because player is no longer running', 2)
            return False, False
```

dagwieers: As I said, this code doesn't read well, we are calling functions with information about both scenarios and then handling each individually in that function where it clearly should be two different cases throughout the code. That would make things more comprehensible.

But in this case I would not be expecting this function to be called when neither `showing_next_up_page` nor `showing_still_watching_page` is set.

resources/lib/stillwatching.py
thebertster: Fair point. Will fix this.

resources/lib/playbackmanager.py
dagwieers: @thebertster Right, I was just going over the code and came to the same conclusion. And might this be related to #139 as well?

I was looking at:
```python
        while (self.player.isPlaying() and (total_time - play_time > 1)
               and not next_up_page.is_cancel() and not next_up_page.is_watch_now()
               and not still_watching_page.is_still_watching() and not still_watching_page.is_cancel()):
            try:
                play_time = self.player.getTime()
                total_time = self.player.getTotalTime()
            except RuntimeError:
                if showing_next_up_page:
                    next_up_page.close()
                    showing_next_up_page = False
                if showing_still_watching_page:
                    still_watching_page.close()
                    showing_still_watching_page = False
                break
```

Personally I think we need to make a different decision in this case. It would be much better if playbackmanager was more closely integrated with `Player()` so we can also handle `onPlayBackEnded` properly and handle it to play the next episode if it was configured this way. @im85288 you wrote this code, what do you think?

dagwieers: @thebertster Maybe if you would add a note to that else-statement, I could agree with this workaround.
```suggestion
            # FIXME: This is a workaround until we handle this better (see comments in #142)
            return False, False
```

thebertster: I don't believe this is related to #139, as we can see the 'playing media episode' message being logged which is at the end of launch_popup() - we then just make the RPC call to play the media and I can't see any way that the playback itself would be aborted. I'm pretty sure this is a Kodi issue where a playback request occurs while the player is in the middle of shutting down...

Personally, I think PlaybackManager should be refactored so that the two pop-ups are handled as separate cases from the start, i.e. decide which of the two pop-ups needs to be displayed at the beginning of launch_popup() based on self.state.played_in_a_row and then branch to handle the two scenarios. There is still a lot of common code in show_popup_and_wait() that can be reused, but it seems ineligant to handle the two cases multiple times through the code.

I think it would make a lot of sense to let onPlaybackEnded trigger the RPC call to Kodi to play the next item. Then we can be sure the previous video player closed down fully. So instead of explicitly playing the next item, just always skip to the end of the current item, setting internal state as to what next item should be played when onPlaybackEnded is triggered (or if it is a playlist, Kodi should just play the next playlist item itself, right?).

However, if the event handling is properly multi-threaded, we could always have a potential race condition between us waiting for the end of the video and the onPlaybackEnded event firing if playback ends between us detecting the end of the video and setting the internal state as to what to play next. We'd need some careful synchronisation to prevent that in the worst case.

dagwieers: Agreed in full. It's not easy, we have gone through this in the VRT NU add-on in order to get resumepoints updated in the cloud and it is messy. It is also something that is hard to build tests around to ensure it works in different scenarios (and e.g. different Kodi versions).

im85288:
I'll be having a look in a couple of days as finally have some time of work! Thanks for the review :)

dagwieers:
@thebertster Can you rebase your branch so you pull in the Travis fixes? I think it would be flagged green after pushing the rebased branch.

thebertster:
Rebase done - Travis checks all passed now.

im85288:
Looks good thanks

PR #144, opened 2019-12-30
Report pylint useless suppressions

dagwieers:
This ensures that we are not suppressing things that aren't being
reaised to begin with (i.e. because we fixed the underlying cause).


im85288:
@dagwieers Looks good to me, ready for merging?

dagwieers:
Yes.

PR #146, opened 2020-01-11
Add notification time offset

CastagnaIT:
This add the possibility to use directly the time offset of credits instead of relying from the final seconds

this should be resolve this issue: https://github.com/CastagnaIT/plugin.video.netflix/issues/414

i made a test and seem solved,
but is better if someother can test, @dagwieers ?

To test is needed to use netflix addon modified from this PR: https://github.com/CastagnaIT/plugin.video.netflix/pull/424

### DIFF
--- a/resources/lib/api.py
+++ b/resources/lib/api.py
@@ -80,6 +80,9 @@ class Api:
         if self.data.get('notification_time'):
             return int(self.data.get('notification_time'))
 
+        if self.data.get('notification_time_offset'):
+            return total_time - int(self.data.get('notification_time_offset'))
+
         # Use a customized notification time, when configured
         if bool(get_setting('customAutoPlayTime') == 'true') and total_time:
             if total_time > 60 * 60:


dagwieers:
@CastagnaIT I am surprised you need changes to **Up Next** for this. IMO **Netflix** can provide the correct `notification_time` without any changes to **Up Next**. Simply calculate the offset from the real runtime of the video and send us that value.

I am not in favor to give 2 mutual exclusive interfaces for the same thing. Unneeded complexity.

CastagnaIT:
In netflix addon dont have this value directly, the only solutions are:
- import Player() to get getTotalTime() and recalculate notification_time. Here means that i should load that module just for this purpose
- use the use the existing json_rpc call to "Player.GetProperties" by adding "totaltime" property, so get the totaltime value and recalculate it to get a total time in seconds (becouse is a splitted time h,m,s), then recalculate again to get a right notification_time. Here there is a lot of calculations

since you already have the total time here, you only perform a simple subtraction calculation instead of all the above mentioned operations

if add this additional possibility is too expensive, there is no problem, i will use one of the my alternatives at the cost of weighing down the operations in my addon already already a big beast compared to this..

im85288:
So Netflix has changed their metadata and the notification time is no longer correct? 

If this is the case and as you mention it is not simple for the Netflix addon to provide the correct notification time I am happy enough to provide the additional offeset option.

But I will leave this for @dagwieers to decide if it makes sense

dagwieers:
I still don't think should be needed. The value we expect to get is the number of seconds before the end of the video. We don't add interfaces for every consumer that has a special quirk.

Looking at the Netflix code, `xbmc.Player()` is used in at least 5 different places. In 4 cases it is just to get the state or a value from the Kodi player. In one case the object is even stored in a class. So it is just an implementation detail of the add-on.

IMO the `services/playback/controller.py` should probably store the `Player()` instance for `section_skipping`, `resume_manager`, `continuity_manager` and `upnext` to reuse. But if not, each call their own instance (as is already the case for 3 of them) when they need information from it. It is easy to fix and in general better to improve this in Netflix.

dagwieers:
@CastagnaIT It seems Netflix is sending the notification to UpNext before the actual media has started. This could be problematic as it will depend on timing whether Up Next will work correctly.

It makes more sense to start playback first, and then after some time (i.e. 1 second) to send the notification. I wanted to implement what is needed in **plugin.video.netflix** to show how you could reduce the number of `xbmc.Player()` instances, while calculating the `notification_time` but this peculiar ordering of things in Netflix may cause issues.

CastagnaIT:
> IMO the services/playback/controller.py should probably store the Player() instance for section_skipping, resume_manager, continuity_manager and upnext to reuse. But if not, each call their own instance (as is already the case for 3 of them) when they need information from it. It is easy to fix and in general better to improve this in Netflix.

Honestly, it's been a while since I've looked at the code on those modules, lately I've been focusing on other issues...
I'm gonna give it a try to unify the "Player" as suggested

to now i close this PR, if i will have problems i let you know

dagwieers:
Let me share what I already did in my branch...

CastagnaIT:
i have see,
but you can move this:
https://github.com/dagwieers/plugin.video.netflix/commit/1d1dd19a1dc08af8a28d5b9712214885939fc8c6#diff-92c555616f72837604676f9bf0140ecfR181

To upnext:

CastagnaIT:
why `except KeyError:` ?
https://github.com/dagwieers/plugin.video.netflix/commit/1d1dd19a1dc08af8a28d5b9712214885939fc8c6#diff-92c555616f72837604676f9bf0140ecfR182

did you find cases where it's missing?

dagwieers:
> but you can move [this](dagwieers/plugin.video.netflix@1d1dd19#diff-92c555616f72837604676f9bf0140ecfR181) to [upnext:](dagwieers/plugin.video.netflix@1d1dd19#diff-0523b6e42738da7adea5d5d2fd7ea128R37] so avoid to recall the player

The call to `xbmc.Player()` should not be a problem, as I said you already do 6 calls to `xbmc.Player()` out of the blue elsewhere in your code. The bigger problem with the Netflix code is related to timing. Since there is no long-lived `Player()` instances the Netflix code cannot use any of the callback handlers. Which is why the code is using `xbmc.sleep()` calls hoping events have happened. Most of these calls to `xbmc.Player()` should happen only after `onAVStarted()`.

I think there is a lot of merit to reorganize the Netflix code, not just for being able to call `getTotalTime()` when you need it, but also to get rid of these `xbmc.sleep()` calls. But given how complex that would be, I concede :-) Let's add the infrastructure you need.

Now rather than call it `notification_time` (which is a misnomer IMO) I would rather call this `credits_offset`. As that is what it is. It is an absolute offset when the credits start.

dagwieers:
> why except KeyError: ? did you find cases where it's missing?

You tell me, you use to have `if 'creditsOffset' in metadata[0]:`, which is there for the same reason.

> in case of RuntimeError: a fall back is possible to the old method

The RuntimeError happens when the `xbmc.Player()` is not (no longer) playing a file. In that case we actually don't want anything to happen. But in your code there is no way to halt sending the notification to Up Next in this case.

dagwieers:
I wanted to modify this existing PR, but I do not seem to have the proper rights for that. I will open a new one in your name.

CastagnaIT:
> > why except KeyError: ? did you find cases where it's missing?
> 
> You tell me 😄 The original code used to have `if 'creditsOffset' in metadata[0]:`, which is there for the same reason.

in that case, I'll take a whipping😶

PR #147, opened 2020-01-13
Add `notification_offset` for Netflix

dagwieers:
Since this matches with Netflix's metadata better and changing the Netflix add-on is too hard at this time.

In the future it is likely other content providers may add this information in the same format as well.

This reimplements PR #146 since I couldn't modify the original PR.

### DIFF
--- a/resources/lib/api.py
+++ b/resources/lib/api.py
@@ -80,8 +80,12 @@ class Api:
         if self.data.get('notification_time'):
             return int(self.data.get('notification_time'))
 
+        # Some consumers send the offset when the credits start (e.g. Netflix)
+        if total_time and self.data.get('notification_offset'):
+            return total_time - int(self.data.get('notification_offset'))
+
         # Use a customized notification time, when configured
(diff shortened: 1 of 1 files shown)


dagwieers:
@im85288 I have conceded :smile: Please merge this.

dagwieers:
With this merged I can see a need to get this out to the Netflix people. Not sure what we want to do with #133, use the straightforward implementation just for now and improve on it later. Or postpone for a next release?

cc @CastagnaIT

CastagnaIT:
oh all right then i'm going back on this road..

if you want to provide the initial functionality for play_url, that's fine for me,
the first time i tried it, it worked like a missile
but now i don't know i see the code's changed a bit

in any way i will wait for your next release after the implementation

dagwieers:
@CastagnaIT The PR has changed a lot, but unfinished. The original implementation was pretty simple, I would be releasing only the simple implementation, not all the other attempts to make it better.

PR #153, opened 2020-01-14
Add System.Time and endtime to notification

### LINKED ISSUE #81: Request : Episode duration / Ends time
Hello !
In my opinion ,that could be a great upgrade to add to the fancy panel the duration of the next episode or even better the ends time (Ends At : 8:00am). 
If this could be possible to integer this feature to a future version that would be awesome. 
Thanks !

dagwieers:
This PR includes:
- Changes to the skin to add Time and next endtime
- New `endtime` window property

This fixes #81 

The implementation looks like this:
screenshot023

### DIFF
--- a/resources/language/resource.language.de_de/strings.po
+++ b/resources/language/resource.language.de_de/strings.po
@@ -27,6 +27,10 @@ msgctxt "#30008"
 msgid "Up Next"
 msgstr "Als Nächstes"
 
+msgctxt "#30009"
+msgid "Ends at: $INFO[Window.Property(endtime)]"
+msgstr "Endet um: $INFO[Window.Property(endtime)]"
+
 msgctxt "#30010"
 msgid "Continue watching"
 msgstr "Wiedergabe fortsetzen"
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -26,6 +26,10 @@ msgctxt "#30008"
 msgid "Up Next"
 msgstr ""
 
+msgctxt "#30009"
+msgid "Ends at: $INFO[Window.Property(endtime)]"
+msgstr ""
+
 msgctxt "#30010"
 msgid "Continue watching"
 msgstr ""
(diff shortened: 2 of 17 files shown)


dagwieers:
So I am wondering whether we should be adding this to the v1.1.0 release as well.
The implementation is rather simple and quite efficient. I was surprised how easy this was.

im85288:
Looks great, let's squeeze it in

notoco:
Does this work with Netflix's library? Somehow I can't get information about the end - both in Netflix and through the library imported from Netflix...

dagwieers:
At least the provided runtime from Netflix's metadata (in Europe?) is off by a few minutes due to translation credits. So it may not be exact.

Not sure if runtime is provided by Netflix add-on and cannot very at this time. @CastagnaIT ?

CastagnaIT:
as I explained in the issue, the problem on netflix is that the video played on the website,
is not the same as the one reproduced on Kodi.

CastagnaIT:
as i explained in the issue, the problem on netflix is that the video played on the website,
is not the same as the one reproduced on Kodi.

In some cases the same video played on Kodi, is missing af a part of the final credits that on the website are viusualized after the final logo of netflix.

for this reason the length of the metadata is wrong, but not because it is really wrong, but because there is obviously something wrong with the way we request the videos

so for the moment it's not reliable,
and now I have no idea how to fix it

notoco:
And all clear. Thanks.

dagwieers:
> as i explained in the issue, the problem on netflix is that the video played on the website,
is not the same as the one reproduced on Kodi.

That is not really relevant here, the question was if the runtime of the next episode was provided to Up Next. As @notoco mentioned this information was missing on-screen.

CastagnaIT:
yes,
i should be able to add a 'runtime' value without any problem

but maybe i found an alternative solution to get the right runtime value, i am not 100% sure
only in theory, i will have to do some tests
i will let you know

CastagnaIT:
ok the alternative solution do not work

@notoco i have added the runtime value anyway
it could be wrong in some cases by a few minutes for the problem explained before,
but even if it's not perfect at least there is

If you can test with: https://github.com/CastagnaIT/plugin.video.netflix/pull/442
let me know

notoco:
Works!

PR #166, opened 2020-01-20
Do not show "Ends at:" when endtime is missing

dagwieers:
This PR includes:
- Avoid showing "Ends at:" when no endtime was available
- Do not strip leading zeroes as "0:34" becomes ":34"
- Small update to changelog (French was missing)


dagwieers:
This is an easy fix, and probably one that requires a new minor release (v1.1.1), but let's not rush into it. We might see more reports the coming days that could go into v1.1.1.

PR #172, opened 2020-02-17
fixed incorrectly formatted endtime

### LINKED ISSUE #169: Window.Property(endtime) wrong
Just noticed it a few time that Window.Property(endtime) is populated wrong

Example:
Correct -> 21:15 
Window.Property(endtime) -> 2121:15

It looks like the hours are duplicated (hh+hh+ss)

Maven85:
This PR fixes #169 

**Background**
https://github.com/xbmc/xbmc/issues/15862

### DIFF
--- a/resources/lib/utils.py
+++ b/resources/lib/utils.py
@@ -149,5 +149,5 @@ def localize(string_id):
 
 def localize_time(time):
     """Localize time format"""
-    time_format = getRegion('time').replace(':%S', '')  # Strip off seconds
+    time_format = getRegion('time').replace('%H%H', '%H').replace(':%S', '')  # Strip off seconds
     return time.strftime(time_format)


resources/lib/utils.py
```diff
@@ -149,5 +149,5 @@ def localize(string_id):
 
 def localize_time(time):
     """Localize time format"""
-    time_format = getRegion('time').replace(':%S', '')  # Strip off seconds
+    time_format = getRegion('time').replace('%H%H', '%H').replace(':%S', '')  # Strip off seconds
```
dagwieers: I am concerned.
In what situation would that ever be %H%H where this would not be a problem otherwise?

I need to understand this before considering any solutions. It might as well be a bug upstream.

resources/lib/utils.py
```diff
@@ -149,5 +149,5 @@ def localize(string_id):
...
     return time.strftime(time_format)
```
dagwieers: I am concerned.
In what situation would that ever be %H%H where this would not be a problem otherwise?

I need to understand this before considering any solutions. It might as well be a bug upstream.

dagwieers: Ok I read the bugreport now 😄

dagwieers:
IMO this is a bug that should be fixed upstream first and foremost. Even in Kodi v18.6.
We shouldn't be delaying backporting this bugfix. The faster this gets backported, the better.

And this PR should also provide context to why this is need and when it can be removed eventually.

dagwieers:
No, don't close it. I want to have fix in Up Next, but it needs some comments to why we do this, with a link to the original issue. So in the future we can remove it when it is no longer relevant.

dagwieers:
Can you reopen this?

Maven85:
Sorry, I deleted my branch. Therefore I cannot open this PR again.

Should I create a new PR?

im85288:
@Maven85  - Yes just create a new PR

Maven85:
I created a PR for Kodi.

https://github.com/xbmc/xbmc/pull/17380

PR #175, opened 2020-02-22
Restore "Ignore Playlist" option

BrutuZ:
Restore the option to not show the pop-up when playing from a playlist.
This prevents a couple issues:
1. Breaking the playlist if the current show has further episodes.
2. Episodes continue playing regardless of the user choice to not continue watching on the pop-ups.

### DIFF
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -26,8 +26,12 @@ class PlaybackManager:
         ulog(msg, name=self.__class__.__name__, level=level)
 
     def launch_up_next(self):
-        playlist_item = True
+        playlist_item = True if get_setting('enablePlaylist') == "True" else False
         episode = self.play_item.get_next()
+        self.log('Playlist setting: %s' % playlist_item)
+        if episode and not playlist_item:
+            self.log('Playlist integration disabled', 2)
+            return
         if not episode:
             playlist_item = False
             episode = self.play_item.get_episode()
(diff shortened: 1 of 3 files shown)


dagwieers:
cc @im85288 @angelblue05 Please review and discuss.

dagwieers:
I would like to release v1.1.1 next week, ideally with this included.
But since this touches the playlist implementation, I need to have feedback from @angelblue05 or @im85288.

im85288:
@dagwieers - I do not use the playlist functionality but after reviewing the code it looks fine to me. 

Unfortunately there are some failed checks but at least no conflicts...so should we merge and get a test release out?

dagwieers:
@im85288 I would like to get all open issues fixed (if possible) before doing a call out to stakeholders for more testing and translations.

resources/lib/playbackmanager.py
```diff
@@ -26,8 +26,12 @@ def log(self, msg, level=2):
         ulog(msg, name=self.__class__.__name__, level=level)
 
     def launch_up_next(self):
-        playlist_item = True
+        playlist_item = True if get_setting('enablePlaylist') == "True" else False
```
BrutuZ: This line failed CI because pylint doesn't like the inline if-test, however just using `bool()` is not an option since the function returns a string (`"True"` or `"False"`) which always evaluates to `True`.
There used to be a similar function that would check for a `.bool` suffix added to the requested setting's name and return a Boolean value instead of a String

BrutuZ:
@im85288 Fixed the the couple grievances CI was pointing out, checks should be passing now 👍

im85288:
@BrutuZ Cool, merged now 😊

PR #177, opened 2020-03-28
Migrate from Travis CI to GitHub CI

dagwieers:
This PR includes:
- Migrate from Travis CI to GitHub CI
- Move the tests from test/ to the standard tests/ directory
- Exclude non-essential files from export
- Move codecov.yml to the only remaining location
- Add Kodi's add-on checker to GitHub CI

PR #178, opened 2020-03-28
Replace screenshots with valid alternatives

dagwieers:
So there is a problem with the currently used screenshots according to
the **kodi-addon-checker**. Since we are now testing much better, we need to
comply with all the rules at any time.

```
ERROR: screenshot should have either 1280x720 or 1920x1080 but it has 1920x1200
ERROR: screenshot should have either 1280x720 or 1920x1080 but it has 1920x1200
ERROR: screenshot is too large 1162KB, maximum file size of 750KB.
ERROR: screenshot should have either 1280x720 or 1920x1080 but it has 1920x1200
ERROR: screenshot should have either 1280x720 or 1920x1080 but it has 1920x1200
ERROR: screenshot is too large 887KB, maximum file size of 750KB.
ERROR: We found 6 problems and 0 warnings, please check the logfile.
```

This fixes the screenshots so we can start using GitHub CI (#177)


dagwieers:
If we merge this one first, #177 will show green when rebased and can be merged after.

dagwieers:
Better screenshots are welcome too. But they have to be small, and use the right dimensions.

im85288:
Cool thanks for sorting this out

PR #184, opened 2020-05-09
Avoid conflict with external players

BrutuZ:
Kodi can't track playback position for external players, so not only we can't tell when to show the pop-up, but when it is displayed (I'm guessing based on `playcountminimumtime` tag from `playercorefactory.xml` - [ref](https://kodi.wiki/view/External_players)) the external player might lose focus and stop playback (was my experience at least, ymmv).

### DIFF
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -27,6 +27,7 @@ class PlaybackManager:
 
     def launch_up_next(self):
         if self.player.isExternalPlayer():
+            self.log('Error: External Player detected...exiting', 1)
             return
         playlist_item = bool(get_setting('enablePlaylist') == 'true')
         episode = self.play_item.get_next()
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -27,7 +27,7 @@ class PlaybackManager:
 
     def launch_up_next(self):
         if self.player.isExternalPlayer():
-            self.log('Error: External Player detected...exiting', 1)
+            self.log('Error: External player detected...exiting', 1)
             return
         playlist_item = bool(get_setting('enablePlaylist') == 'true')
         episode = self.play_item.get_next()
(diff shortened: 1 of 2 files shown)


resources/lib/playbackmanager.py
```diff
@@ -26,6 +26,8 @@ def log(self, msg, level=2):
         ulog(msg, name=self.__class__.__name__, level=level)
 
     def launch_up_next(self):
+        if self.player.isExternalPlayer():
```
dagwieers, line 29: We would need to log when this happens otherwise it will be harder to troubleshoot issues caused by an external player.

dagwieers:
I fixed two issues causing this PR to fail.
If you add logging to this PR, the re-run tests will work fine.

dagwieers:
You still have to rebase your branch (`git pull --rebase origin master`).

BrutuZ:
#185 wasn't merged yet, that's why I'm having trouble with Fetch 😅

dagwieers:
> #185 wasn't merged yet, that's why I'm having trouble with Fetch sweat_smile

Ouch, my bad :roll_eyes:

dagwieers:
So I am wondering, how do you enable an external player? Does it always require a Kodi restart?

BrutuZ:
Kodi doesn't restart, it's just sent to the background while the other player takes care of the media playback.
In my case I had a few files that wouldn't play natively but MPC had no problem playing them. So using the example referenced above, I set it up to auto-play in full-screen and exit when playback is finished (not all players have that last setting which makes it basically seamless) so apart from the few milliseconds of splash-screen while it loaded the file and lack of Kodi's player UI, it all feels the same since it's back to Kodi once playback is over.
---
That's reasonable, I'll move it then. I didn't go through the entire code, I happened to have that file open from previous edits and this just happened to work there too.

dagwieers:
Thanks !

PR #186, opened 2020-05-10
Prepare for v1.1.1 release

dagwieers:
Update README and addon.xml for release.


dagwieers:
@im85288 @mediaminister I think a new release is needed, we have fixed a few known issues, but never released them to our users. I will call out to the translators.

mediaminister:
Okay, I tested the master branch on Linux and Windows in Kodi 18 and Kodi 19 and it works.

dagwieers:
All my tests worked fine as well, despite that we have a few reported issues that we could not pinpoint yet. Let's wait a few days to see if we get more translations, or feedback on the open issues.

PR #192, opened 2020-06-22
Add check for abortRequested

dagwieers:
This was requested by Kodi devs.

### DIFF
--- a/resources/lib/script.py
+++ b/resources/lib/script.py
@@ -5,7 +5,7 @@
 from __future__ import absolute_import, division, unicode_literals
 from datetime import datetime, timedelta
 from math import ceil
-from xbmc import Monitor
+from xbmc import Monitor, sleep
 from xbmcgui import WindowXMLDialog
 from statichelper import from_unicode
 from utils import addon_path, get_setting_bool, localize, localize_time
@@ -93,7 +93,7 @@ def test_popup(window):
     while popup and step < timeout and not monitor.abortRequested():
         if popup.pause:
             continue
-        monitor.waitForAbort(wait)
+        sleep(wait)
         popup.update_progress_control(timeout, wait)
         step += wait
 


im85288:
Sorry I already committed my changes! They also wanted the sleep out, it'll make things easier just to give them what they want rather than risk another round of tag/untag etc just for something we only use for testing creating skins

im85288:
How about this from @anxdpanic:

```
monitor = Monitor()
while popup and step < timeout and not monitor.abortRequested():
	if popup.pause:
		continue
	if monitor.waitForAbort(wait):
		break
	popup.update_progress_control(timeout, wait)
	step += wait
```

dagwieers:
It is always better to use PRs and use a review process. At least the PR gets sanity and unit tested (unit tests up to 42%)

dagwieers:
Again, I don't think it makes sense to make it more complicated than it has to be. Using sleep is fine for values of 100ms, and I expect the overhead to be bigger when using `waitForAbort()`. There is no need to abort faster than 100ms. If that would be the case, we have other code paths where we are using sleep() now.

im85288:
Ok cool let's go for this and see what they say.

dagwieers:
I rebased my PR.

dagwieers:
We probably want to test this for a few hours before putting this up again.

im85288:
Too late :) I submitted them for a review. Let's see what happens and test in the meantime. Strangely enough for Leia the Travis build worked and for Matrix it failed!

dagwieers:
The downloads are loadbalanced and apparently downloads.kodi.tv has an expired SSL certificate :-/

im85288:
Oops! Someone needs to do some configuration magic.

dagwieers:
I actually implemented HTTP retries to recover from issues like this: https://github.com/xbmc/addon-check/pull/217
So I don't understand why this did not help here. I can see it retried 5 times, but probably the same host.

urllib3 should support round-robin DNS entries though.

im85288:
The submit was approved this time. In matrix repo in a couple of hours apparently. Cool

im85288:
Also for future after a review we should Only need to remove the tag then retag.
Closing the PR is not needed:


git tag -d <tag_name>

git push --delete origin tagname

git tag tagname && git push --tags

dagwieers:
Now let's hope we did not break it :wink:

im85288:
Fingers crossed!

dagwieers:
I tried the v1.1.1 release and I do not see a pop-up appearing :disappointed:

im85288:
Oops, it must be that waitForAbort thing in player. Probably should have tested it first!

dagwieers:
I changed it back to a sleep and things work correctly again. But I do not understand what is going on.

im85288:
I wonder if it needs to be conditional instead? Looking at the implementation it seems it returns no matter what so yes very strange

dagwieers:
Ok, so waitForAbort() is in seconds, where sleep() is in ms :roll_eyes:

dagwieers:
I guess we need v1.1.2 now...

im85288:
Whoops! Yes I guess we do 😊

im85288:
Corrected this now, will push out v1.1.2

Issue #170, opened 2020-02-12
Not playing next item


prydda:
Using Embycon to playback and embuary skin. 
I'm getting the popup to play next episode but it never does, it jumps back to current season. 

logfile: http://ix.io/2bxL

dagwieers:
The log is mostly empty. Did you enable debug logging?

prydda:
Nope, sorry. 
here is with debug activated: 
https://del.dog/ygarfullyg

dagwieers:
I don't know who wrote this dogbin service, but it is unusable in Chrome. Is it mining bitcoins perhaps?

How hard is it to create a service to show a text file in a browser without bringing the browser to a grinding halt. Sorry, try again.

dagwieers:
Maybe you can attached it to this issue instead?

prydda:
Not sure, used it for android logs. ix.io didn't work - I beileve the file was to large? (5.4mb) 

here is a github gist: https://gist.github.com/prydda/4cd7c341cb0d235a8a0f49c621525435

dagwieers:
You can also attach it, see below: *Attach files by dragging & dropping, selecting or pasting them.*

dagwieers:
The only logged information from Up Next is receiving the information:
[...]

PR #195, opened 2020-06-30
Automatically take action after X seconds (into credits)

behunin:
Give an option to automatically play the next item or stop the player, with control over how many seconds before action is taken.

### DIFF
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -132,12 +132,24 @@ msgctxt "#30607"
 msgid "Number of episodes before \"Still there?\" query"
 msgstr ""
 
+msgctxt "#30608"
+msgid "When credits offset is known (e.g. Netflix)"
+msgstr ""
+
 msgctxt "#30609"
 msgid "Duration in seconds for notification to be shown"
 msgstr ""
 
+msgctxt "#30610"
+msgid "Duration for automatic default action"
+msgstr ""
+
 msgctxt "#30611"
-msgid "Customize autoplay duration by episode length"
+msgid "Customize notification time by episode length"
+msgstr ""
+
+msgctxt "#30612"
+msgid "When credits offset is unknown"
 msgstr ""
 
 msgctxt "#30613"
--- a/resources/language/resource.language.en_gb/strings.po
+++ b/resources/language/resource.language.en_gb/strings.po
@@ -133,11 +133,7 @@ msgid "Number of episodes before \"Still there?\" query"
 msgstr ""
 
 msgctxt "#30608"
-<<<<<<< HEAD
 msgid "When credits offset is known (e.g. Netflix)"
-=======
-msgid "Automatic Watch or Stop"
->>>>>>> 02eeee67b0675df4b98fd39fa11ad25295e303c9
 msgstr ""
 
 msgctxt "#30609"
@@ -145,11 +141,7 @@ msgid "Duration in seconds for notification to be shown"
 msgstr ""
 
 msgctxt "#30610"
-<<<<<<< HEAD
 msgid "Duration for automatic default action"
-=======
-msgid "Duration in seconds for auto watch or stop"
->>>>>>> 02eeee67b0675df4b98fd39fa11ad25295e303c9
 msgstr ""
 
 msgctxt "#30611"
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -96,9 +96,13 @@ class PlaybackManager:
         except RuntimeError:
             self.log('exit early because player is no longer running', 2)
             return False, False
+        if self.api.has_addon_data().get('notification_offset'):
+            next_up_progress_step_size = calculate_progress_steps(get_setting_int('autoTime'))
+        else:
+            next_up_progress_step_size = calculate_progress_steps(total_time - play_time)
         progress_step_size = calculate_progress_steps(total_time - play_time)
         next_up_page.set_item(episode)
-        next_up_page.set_progress_step_size(progress_step_size)
+        next_up_page.set_progress_step_size(next_up_progress_step_size)
         still_watching_page.set_item(episode)
         still_watching_page.set_progress_step_size(progress_step_size)
         played_in_a_row_number = get_setting_int('playedInARow')
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -96,11 +96,7 @@ class PlaybackManager:
         except RuntimeError:
             self.log('exit early because player is no longer running', 2)
             return False, False
-<<<<<<< HEAD
         if self.api.has_addon_data().get('notification_offset'):
-=======
-        if get_setting_bool('autoWatchOrStop'):
->>>>>>> 02eeee67b0675df4b98fd39fa11ad25295e303c9
             next_up_progress_step_size = calculate_progress_steps(get_setting_int('autoTime'))
         else:
             next_up_progress_step_size = calculate_progress_steps(total_time - play_time)
--- a/resources/lib/upnext.py
+++ b/resources/lib/upnext.py
@@ -92,18 +92,10 @@ class UpNext(WindowXMLDialog):
         if runtime:
             self.setProperty('endtime', from_unicode(localize_time(datetime.now() + timedelta(seconds=runtime))))
         if self.current_progress_percent <= 0:
-<<<<<<< HEAD
             if get_setting_int('autoPlayMode') == 0:
                 self.set_watch_now(True)
             else:
                 Player().stop()
-=======
-            if get_setting_bool('autoWatchOrStop'):
-                if get_setting_int('autoPlayMode') == 0:
-                    self.set_watch_now(True)
-                else:
-                    Player().stop()
->>>>>>> 02eeee67b0675df4b98fd39fa11ad25295e303c9
 
     def set_cancel(self, cancel):
         self.cancel = cancel
--- a/resources/lib/upnext.py
+++ b/resources/lib/upnext.py
@@ -7,7 +7,7 @@ from platform import machine
 from xbmc import Player
 from xbmcgui import WindowXMLDialog
 from statichelper import from_unicode
-from utils import get_setting_bool, localize, localize_time
+from utils import get_setting_int, get_setting_bool, localize, localize_time
 
 ACTION_PLAYER_STOP = 13
 ACTION_NAV_BACK = 92
@@ -91,6 +91,11 @@ class UpNext(WindowXMLDialog):
             self.setProperty('remaining', from_unicode('%02d' % remaining))
         if runtime:
             self.setProperty('endtime', from_unicode(localize_time(datetime.now() + timedelta(seconds=runtime))))
+        if self.current_progress_percent <= 0:
+            if get_setting_int('autoPlayMode') == 0:
+                self.set_watch_now(True)
+            else:
+                Player().stop()
 
     def set_cancel(self, cancel):
         self.cancel = cancel
--- a/resources/settings.xml
+++ b/resources/settings.xml
@@ -9,7 +9,9 @@
         <setting label="30032" type="bool" id="enablePlaylist" default="false"/>
         <setting label="30605" type="bool" id="includeWatched" default="false"/>
         <setting label="30607" type="slider" id="playedInARow" default="3" range="0,1,15" option="int"/>
-        <setting label="30609" type="lsep"/> <!-- Autoplay duration -->
+        <setting label="30608" type="lsep"/>
+        <setting label="30610" type="slider" id="autoTime" default="6" range="0,3,120"/>
+        <setting label="30612" type="lsep"/> <!-- Autoplay duration -->
         <setting label="30611" type="bool" id="customAutoPlayTime" default="false"/>
         <setting label="30613" type="slider" id="autoPlaySeasonTime" default="30" range="0,5,120" option="int" subsetting="true" visible="eq(-1,false)"/>
         <setting label="30615" type="slider" id="autoPlayTimeXS" default="15" range="0,5,120" option="int" subsetting="true" visible="eq(-2,true)"/>
--- a/resources/settings.xml
+++ b/resources/settings.xml
@@ -9,15 +9,9 @@
         <setting label="30032" type="bool" id="enablePlaylist" default="false"/>
         <setting label="30605" type="bool" id="includeWatched" default="false"/>
         <setting label="30607" type="slider" id="playedInARow" default="3" range="0,1,15" option="int"/>
-<<<<<<< HEAD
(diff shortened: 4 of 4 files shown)


im85288:
Thanks, I am going to see what @dagwieers thinks about this functionality before pulling in.

dagwieers:
Sorry for keeping everyone waiting. When I saw this PR I wasn't sure what this adds as we already have a default action (play next or stop playback). So a bit of information on the use-case is welcome.

behunin:
I use this addon mostly with the netflix addon and wanted to be able to have the next episode play within a few seconds of the end credits, much in the same way netflix does with the player on their site. I decided it would be nice to have control over how many seconds before action was taken and added that as well.

Before this PR I would see the progress bar go to 0% and then see the countdown start with no automatic action taken.

I am on the latest commit of both kodi and this addon.

dagwieers:
@behunin I am confused. You first see the progress bar go to 0%, and *then* the countdown start? That seems like a bug. Countdown and progress bar should be simultaneous. If you can reproduce this, please open a new issue for this.

The default action (autoPlayMode) is configured in the settings, and could be "Play next" or "Stop playing". It defaults to "0" (which is "Play next"). If it doesn't automatically play next after the countdown, then that is a bug. (#139 and #170 were reported before, and #198 proposes a fix)

Having a system-wide option to automatically play (i.e. not wait for a countdown counter) is something I don't think is very useful. It will fail for anything but Up Next. (local content or other VOD add-ons do not have an exact timing for the credits and therefore could cut off the episode prematurely).

So if anything, I think it should be an option to Netflix (or any add-on where this use-case makes sense) that we could support.

@CastagnaIT Is this something you have had requests for before? Does Netflix support such a feature already?

CastagnaIT:
Hi,
I think he wants to add an option to have a possibility to set a custom timeout time (for automatic behaviour) now is set to 1minute?

personally i do not use upnext but wait a few (customizable) seconds instead of 1 minute i think it's nice

>  much in the same way netflix does with the player on their site

he refer to website, the autoplay of next episode have a very short timeout (some seconds)

> Before this PR I would see the progress bar go to 0% and then see the countdown start with no automatic action taken.

i do not understand this sentence, perhaps he means a bug somewhere?

He also add "Automatic Watch or Stop" option menu
but seem to add again an existing feature
if it is not, is better that @behunin explain the use case and what differs from 
"Default action when nothing select" option menu

behunin:
> You first see the progress bar go to 0%, and then the countdown start?

Yes that is the observed behavior. I was confused by this as well.

This PR checks to see if the UpNext progress bar <= 0 and then takes action if the option I added is set to true.

My intention was to recreate the same functionality as netflix has on its player so maybe this needs to be added to the netflix repo. What are your thoughts on that @CastagnaIT?  

I will investigate further today and gather screenshots for a new issue as requested.

CastagnaIT:
I don't understand anything about what you want to do...
netflix add-on integrates upnext like any other VOD/Streaming add-on
so nothing is specific

im85288:
Yes I am also confused here @behunin perhaps some screenshots will tell the story better. @CastagnaIT - the last time I used the Netflix addon the popup for upnext was not occurring when the credits started to show. The time before that though it was working fine in that as soon as the show ended the popup came...do you know if this functionality (i.e. calling the show notification time) has changed?

CastagnaIT:
> the last time I used the Netflix addon the popup for upnext was not occurring when the credits started to show.

upnext could not works only in this case:
Kodi 19 + play nf videos from Kodi library

Because the JSON RPC Player.GetProperties in some cases is broken and provide broken data that cause unexpected behaviuours of the add-on service, but usually when happens is shown a popup error notification.

if this is not the case, you'll have to give me a case with steps to replicate it by opening an Issue

im85288:
Thanks it may have been Kodi 19, I've been messing with all sorts so not sure. I'll check and create an issue if it's not Kodi 19.

behunin:
I did an uninstall of kodi and then installed kodi 19 @ [`7a65d55`](https://github.com/xbmc/xbmc/commit/7a65d555c412baccf32efe3c23959e280572c165), then installed both netflix and upnext using kodi's addon manager.

The strange behavior I mentioned was no longer present so that was my bad. Sorry...

What I did observer is that with "Customize autoplay duration by episode length" set to true and all lengths set to 5. There was no change in the length of time upnext played the next episode. This must be what brought me to make this change and the confusion on my part.

This PR does work as I intended so I would like to refactor it with the settings I added to be auto play for any length. Then have each setting for auto play time disable the other so that only one can be set true.
So either auto play for any length or auto play for custom lengths can be true, not both. 

What do you gentlemen think about that idea?

dagwieers:
> What I did observer is that with "Customize autoplay duration by episode length" set to true and all lengths set to 5. There was no change in the length of time upnext played the next episode. This must be what brought me to make this change and the confusion on my part.

The "Customize autoplay duration by episode length" influences the amount of seconds before the end of the video the pop-up will appear, *if and only if the video does not already provides this information*. Since Netflix already provides detailed information on when the pop-up needs to be shown (i.e. when the credits start) this has no effect on Netflix. It affects local media and other VOD add-ons only.

A generic option that automatically performs the default action is not foreseen, and I am not sure if Netflix does this currently? If we add this, this would only makes sense for VOD providers that provide detailed information on when the credits start (i.e. Netflix). But I wonder if this would not be something people would configure on a per-addon basis instead.

I am open to the suggestion, but would need to understand how this works in Netflix before making a call.

dagwieers:
I modified the issue subject hoping this is what you intended, if not, feel free to clarify it further :-)

CastagnaIT:
there is no UpNext customization on netflix addon
the start of the next video, is done according to the UpNext behaviour nothing more...

im85288:
Just to clarify, yes there is no customisation but the Netflix addon is the only one (as far as I am aware) that uses the facility to show the Popup at the exact time the credits start.

dagwieers:
@CastagnaIT I understand. But if this functionality is offered by Netflix today, we could make Up Next behave similarly. We have two options for add-ons that have detailed information on credits:
- New option in Up Next settings: **Automatically perform default action after X seconds into credits**
- A new key in the data structure send from the add-on that specifies this value

We could also implement both, a setting in Up Next that can be overridden by the add-on.

My concern is that these options only make sense if we have detailed information on credits, and this is not obvious to the user. So we probably need to reorganize the Up Next settings a bit to make this more clear (no idea yet on how to do this best).

CastagnaIT:
This is the NF website behaviour:

when credits time is fired will show these buttons
image
NF let to user only about 6 seconds to choose what doing
if the user not do a choice automatically fire "Next episode" button

Now i see your problem,
if the credits offset value is not available, this new behaviour option not have much sense

maybe yes, a solution is let an add-on the choice to enable a window time for the default action
but only with a boolean option parameter,
and so let user set the window time length in UpNext settings
but you should warn the user that it only works with add-ons that support this feature

im85288:
Don't we already support this feature? The user can choose if the default action is to 'Play Next' the default and the time in seconds for the notification to be shown (5, 10, 15, 20, 25, 30) etc etc

dagwieers:
Netflix doesn't even offer the option to stop watching, it either is "Continue watching credits" (which in our case is the same as escape/back to dismiss the pop-up, or not doing anything) or "Playing next episode". Obviously you can always stop watching in the next episode, so this is convenient (if you have the credits offset).

So we could implement this behaviour, both visually as well as functionally: Show a "Watch credits" and "Next episode" buttons, but the current settings we have would be even more confusing if we don't clearly differentiate the settings by whether we have a credits-offset or not.

@CastagnaIT I will let you decide because you are the only consumer of this functionality if it were available.
Is this something you would like to offer? And would you make it configurable in Up Next, or set by the add-on?

dagwieers:
@im85288 Yes, we offer a choice, but we don't automatically continue with the default action after 6 seconds, the user either has to grab for the remote, or wait until the end of the credits. It hasn't bothered me, but I can see how this would accommodate the lazy watcher ;-)

im85288:
At yes I get it now, so for all the other addons where we show the popup X seconds from the end it works fine..but yeah I agree for Netflix where sometimes you have to wait 2 minutes or even more this would be a handy feature to build in. In fact I think if we can work it out it should just work that way by default..eg. Does the play next after the countdown finishes or whatever the user has selected as the default Action.

CastagnaIT:
if you want to implement the window time customization for automatic default action it's fine to me
i will activate the functionality when it is ready

I would prefer just to add a boolean property key to the data to send in order to enable it

dagwieers:
@CastagnaIT I was more thinking of a time window in seconds (e.g. autoplay_delay) that would be provided. For Netflix that would then be 6 (seconds). If it would not be provided, we would not act on it. (So not add this as a setting in Up Next even)

But if this would be set, we could also modify the pop-up to only have "Watch credits" and "Play next episode" and maybe even make the "Play next episode" the default action. It would mean that some Up Next settings would no longer affect Netflix (and other add-ons providing credits offset)

I am not 100% sure if we want this (i.e. change the look and feel) for Netflix. It could be confusing for people that do not care if something was played from Netflix or another means.

CastagnaIT:
I don't know who proposed changing the interface to UpNext, but that's not my intention

I'm just saying the time window parameter for "auto play?" or "auto select action" it should be an UpNext settings
not should be handled from add-ons that integrate upnext

I'm not in favor of adding a setting parameter in my add-on settings for something that should be handled "externally"
I don't see the point...

dagwieers:
That's clear, I will not propose to make it more Netflix-alike.

However you proposed [here](https://github.com/im85288/service.upnext/pull/195#issuecomment-658171893) to make it add-on controlled, but I now understand you don't see a need for that. No problem.

CastagnaIT:
perhaps i explained it wrong way

behunin:
Instead of adding any length option, how about a option for action after x seconds on any addon that uses upnext with the notification_offset set. Then make it clearer that the other timing options are for addons that don't use that option.

CastagnaIT:
I personally have no intention of maintaining external add-on features inside the add-on

dagwieers:
@behunin I think your implementation is fine in general. I still need to think on how we can present what options work for which cases to not make things more confusing for our users. Maybe we ought to split the options between: Having detailed credits offset and not having them, but I can't find a good descriptive title for this.

CastagnaIT:
Can you explain what means this option? that i do not understand
image

Dnkhatri:
> Can you explain what means this option? that i do not understand
> image

That setting is for local files and starts upnext widget according to the length of the video file. For netflix addon as it is already known where the credits start it is ignored.

CastagnaIT:
i can suggest something like this:
> "-" == submenu

[group-------------Automatic button action-------------]
Autoselect action (when nothing selected) [Play next/Stop]
Enable custom autoselect time [true/false]
-For calculated credits position [slider/seconds]
-For precise credits position  [slider/seconds]

behunin:
**Automatic Actions after X seconds**

1. Timing for local content, VOD addons, etc
    X seconds for Y length of episode...
2. Timing for addons that send detailed offset information. [Netflix]
    X seconds. 

Number one I end with an etc because I don't know all the situations this applies to.
Number two is my suggestion, with the array of addons that use this offset information growing when the time comes.

dagwieers:
I was thinking of doing it like this.

**Behaviour**

> ***Features***
> Default action when nothing selected
> Include already watched episodes
> Number of episodes before \"Still there?\" query

> ***When credits offset is known (e.g. Netflix)***
> Duration before automatically performing default action

> ***When credits offset is unknown***
> Duration in seconds for notification to be shown
> Customize notification time by episode length

behunin:
So are you saying just X seconds for both? I think just having two timings would make it simpler, no boolean, just a slider for each timing. Default 6 seconds for both?

dagwieers:
@behunin No, it doesn't make sense to have a value if we don't have an exact credits offset. This functionality *only* makes sense when we do. (see https://github.com/im85288/service.upnext/pull/195#issuecomment-658864715)

im85288:
@dagwieers could this also work for items where the credits offset is not known (ie. anything but Netflix at the moment)..for example you want the popup to show 30 seconds from the end, but then you also want it to play the next episode automatically after 6 seconds of no activity..In other words perfect for lazy people?

dagwieers:
@im85288 No, because we don't know if the credits will be at 30 - 6 (24 seconds). It might only be at 10 seconds before the end, and you cut off the video prematurely (see https://github.com/im85288/service.upnext/pull/195#issuecomment-658864715). We already estimated the values to be before the credits start (because if the notification is shown after the credits are shown, that is not convenient either).

im85288:
@dagwieers - I don't follow. I'm only referring to the path when the pop up is shown at the use defined setting. So if it is shown 30 seconds from the end we know there is 30 seconds until it would auto play. In that case we could allow auto play after a period shorter than the 30 seconds. 

I may be missing something??

dagwieers:
@im85288 I am referring to that same path. So what if the credits only start 10 seconds before the end. Notification time is 30 and autoplay time is 6. You are cutting off the last 14 seconds of an episode.

If we don't know when the credits actually start, the following is important:
- The notification time we propose should be before the actual credits start
- The user needs to acknowledge explicitly when to start the next episode

We cannot offer automatic proceeding because we have no clue if/when the credits have started. And we cannot risk cutting of a video prematurely. With EDL support we could offer it for local files though, but IMO EDL support should be part of Kodi's metadata, not an Up Next feature IMO.

im85288:
Fair enough, I understand what you mean. It is probably safer as you suggest not to add this feature then for the occasion when we don't know exactly when the credits will start.

I agree about EDL being preferred as a native kodi feature. Plex has recently started offering the Skip Intro feature on their clients (Apple TV etc) and it works well. They do some sort of analysis on the files prior which is clever stuff.

behunin:
@dagwieers @im85288 Please review revisions.

dagwieers:
@behunin Will check later tonight!

dagwieers:
@behunin Next time you make a PR, please always make a separate branch and dont use your master branch. Because of this I cannot  add the necessary commits to your branch and add the work as a single commit... I will have to make a new PR with your commits included.

Dnkhatri:
> Fair enough, I understand what you mean. It is probably safer as you suggest not to add this feature then for the occasion when we don't know exactly when the credits will start.
> 
> I agree about EDL being preferred as a native kodi feature. Plex has recently started offering the Skip Intro feature on their clients (Apple TV etc) and it works well. They do some sort of analysis on the files prior which is clever stuff.

Yeah it should be a Kodi feature but some of the edl functionality got broken after 17.6 release and has never been fixed till now. I don't think any of the current Kodi dev use this feature much so it has been neglected. As Kodi is a hobby project for most of them so I don't blame them as they would work on things that interest them but I have given up hope that it will get implemented. I waited for 6 months hoping it will get implemented in kodi 19 but with the Alpha to drop and no progress on it I thought I put in my request again.

PR #200, opened 2020-07-15
Automatically take action after X seconds (into credits)

dagwieers:
This PR adds a feature to automatically advance to the next episode after X seconds just as Netflix does.
It is only activated when notification_offset (credits offset) is provided, currently only used by the Netflix add-on.

This replaces #195.


behunin:
Thank you @dagwieers 
Sorry for the headache

behunin:
I was thinking that maybe it would be good to add a boolen in api when notification_offset is used. That way show_popup_and_wait doesn't have to load the api data, just check a bool.

dagwieers:
There is more than one problem with the implementation unfortunately:
- We do not set the correct remaining time, so it no longer shows "Next episode in XX seconds"
- It doesn't support both `notification_time` and `notification_offset`
- We should probably add some supporting code to make it more readable

This is unfortunately a lot harder to do right than I anticipated :disappointed:

behunin:
> It doesn't support both notification_time and notification_offset

This is why I suggested adding a bool to api.
See [ca3ffb](https://github.com/behunin/service.upnext/commit/3fb01527ca3ffba362a05961d0bba35e435f529f)

> We do not set the correct remaining time, so it no longer shows "Next episode in XX seconds"

After reasoning about this for a while I see what you mean by this being not so easy after all.

behunin:
I think most users will leave the default time to be 6 seconds and if we change the max to be 12 then I think a countdown is not needed. This feature is for the binge watcher that doesn't care what the count is, just that it works and the binge continues. 🤣

Issue #203, opened 2020-07-27
Use with "Play next video automatically"


ltGuillaume:
I am using Kodi via LibreELEC on a Raspberry Pi 2, without the use of the library. I'm trying to get Up Next to work when I start playing a single file on an SMB share with several files in the same path.

In the Kodi settings under Player > Videos, there's the option "Play next video automatically", which automatically adds other files in a folder to Kodi's playlist (queue).

I assumed that this is the playlist (queue) that Up Next refers to in its settings ("Enable on playlists"), but am I wrong? Does Up Next only work for saved playlists?

If so, can Up Next also be made to work for Kodi's ad-hoc playlist/the queue?

__Edit__: I may have searched the issues with the wrong keywords, may be duplicate of #199. But trying the latest master (with #198 merged) did not change anything.

dagwieers:
I would expect this to work already. There should not be a difference in how this works for local or SMB/NFS content. (I am using NFS myself).

In all three cases, we rely on Kodi's media library to know what the next episode is based on the existing episode and the information stored in Kodi. We do not need the "Play next video automatically" to be enabled. There is no advantage to have this, if you have Up Next installed.

Please share your Kodi debug log (preferably, restart and perform a minimum set of action to reproduce this) so we can see what Up Next is doing.

ltGuillaume:
Aha, but there is! I do not use the Kodi library in any way: none of the local and SMB locations I have set up are included in the Kodi database. 

As such, Up Next won't be able to find the next episode in the database, but it _would_ be able to honor the next playlist item.

I thought this would be where the setting "Enable on playlists" would come in. I'd say it's rather confusing that it doesn't.

I went through the debug log, but there is no mention at all of Up Next trying to find a file to play next, but that makes sense if it only considers entries from the database.

dagwieers:
So the main question remains why Kodi is not playing the next file. What type of files did you select in the settings? I assume Kodi does not know what is a series or a movie if it did not index the files...

But I think we can close the ticket.

ltGuillaume:
Kodi does add the files to the playlist and it will play the next file when it's at the end of a file. I just thought that "Enable on playlist" would trigger the Up Next popup 30 seconds (or custom time) before the end, because there's a next item in Kodi's playlist. Apparently it doesn't.

ltGuillaume:
> I just thought that "Enable on playlists" would trigger the Up Next popup (...) Apparently it doesn't.

[...]

Issue #199, opened 2020-07-13
Doesn't auto-play next episode


nuentes:
I've been having a weird issue with UpNext for a while. Hard to say how long, but I believe it goes back several versions. The dialog comes up just fine, and I can click both buttons, and they both work correctly if I manually click them. However, if I do nothing, the video simply ends. In the settings, I definitely have "Default action when nothing selected" correctly set to "Play".

I've tried an uninstall/reinstall of UpNext, and I've also downgraded from 1.1.2 to 1.0.5, and I have the exact same issue with either version. I was previously using the BOX skin, but I'm switching to Arctic Zephyr 2 and seeing the same issue on AZ2. All my media is added via Emby for Kodi. I have also confirmed that I do not have anything selected in the "Play next video automatically" setting for Kodi.

This logfile is from a clean Kodi boot (I still have 1.0.5 at the time of testing). I played an episode of a series. I skipped to the end of the episode and waited for the dialog to come up. For the first test episode, I selected "Watch Now" and the next episode correctly played. For the 2nd episode, I simply did nothing. The dialog came up, and the video player simply closed after the 20 seconds. Relevent part of my kodi logfile: https://paste.kodi.tv/vepepereca.kodi

dagwieers:
This is probably a duplicate of #139 or #170, please test PR #198

nuentes:
It looks like those commits were merged into master, so I downloaded and manually installed a zip of the master branch, and still had the same issue. I disabled and re-enabled the service just to be sure. I previously still had 1.0.5 installed (with auto-updates disabled) and now it shows 1.1.2 again.

I didn't mention it before, but this is installed on an Android device, if that's relevant.

nuentes:
Based on updates in 198, it sounds like this is something that could potentially be resolved by the way an addon handles playback. Would it make sense to report this in the Emby forums?

im85288:
@nuentes Emby for Kodi has two playback methods. If you selected Direct Paths (which I advise) then the behaviour is just like having your media locally installed in Kodi. The other method would use our addon integration method for showing upnext. It has all worked well in the past but not sure why it's not now. @angelblue05 does not seem to be active on Emby for Kodi so I think you'll have more success pursuing the issue here.

nuentes:
Direct Paths requires me to change how my media is added in Emby. Right now my media is added via local folder paths. However, in the near future, I hope to move them to NFS or SMB paths, which would allow me to switch to Direct Play.

[...]

Issue #197, opened 2020-07-01
No pop-up first video


hockeymachine86:
When I play a video (local library), I get no pop-up the first time I open kodi to watch.  Meaning that if I open kodi and start watching a series, no pop-up.  But if I manually select the next episode in the series, when I get to the end of that episode, I get an episode.

From the log: 
2020-06-30 20:07:48.629 T:7497  - Start watching episode 
2020-06-30 20:52:29.007 T:7497  - I have to manually open the next episode
Log Part 1: https://pastebin.com/sczn9FPx
Log Part 2: https://pastebin.com/fG2rRt2c

dagwieers:
I think this is because Kodi does not know what the season and episode is because it does not understand the formatting. You probably have to reconfigure Kodi to make it understand 3x19, or rename your files to use the S02E05 convention.



This is not an Up Next bug.

hockeymachine86:
Thanks. Is there a reason it shows up after the first video plays. Meaning if I watch 3x19, it doesn't popup, but if I open 3x20, it will popup when that one ends for me to continue onto 3x21?

dagwieers:
No, it should either work for all, or not work at all. And that is how it behaves for me with local video too. Only when Kodi doesn't know what the season/episode is, or did not determine that the next episode is available it will not show a pop-up notification. In your case it should if both requirements are met.

hockeymachine86:
Okay, so the issue I have is it doesn't work for the first video but then works onward.

hockeymachine86:
Wondering if it's because it doesn't know it exists. Am playing from a samba share. I have experienced the same from shield and firestick for reference.

dagwieers:
And this is the case for every series and every season? Or just in this particular case?
Also, episode 19 is not the first episode, so this is confusing.

dagwieers:
Are you playing from the Series interface in Kodi, or just from a video-listing?
Because, as I said, Kodi needs to know the season/episode of the file being played, and needs to know the next episode exists.
If you play from a video listing, it doesn't know any of this information, unless some files have been processed.

So I still don't think this is an Up Next issue, but related to how you have added files to Kodi. (Some have been indexed, others have not?)

hockeymachine86:
It's playing from a series listing. Yes happens with all series and seasons/episodes. Right episode 19 is the example of last watched episode. Each show is in its own folder and was added to the library as a TV show with each folder containing only a single TV show.

dagwieers:
[...]

Issue #196, opened 2020-07-01
Duplicate title


hockeymachine86:
When UpNext pop-ups up it shows a duplicate banner as seen in the photo attached. This is for playing a local file. 
![20200630_213907](https://user-images.githubusercontent.com/47342202/86272323-01531f00-bb94-11ea-967e-e10a2d01b15a.jpg)

dagwieers:
Please test the upcoming v1.1.3 release by testing PR #217 and reporting back.

PR #204, opened 2020-07-29
Fix queuing of next episode using playlist

thebertster:
Fixes issues with queuing of next episode via playlist:

1. We should not reset state if an episode ends, triggering the onPlayBackEnded event, when we have the next episode queued.
2. We should queue the next episode when the "Up Next" dialog is presented, but dequeue the next episode if the user cancels playback of the next episode.
3. We should not queue the next episode when the "Still Watching?" dialog is presented until and unless the user indicates they are indeed still watching.

### DIFF
--- a/resources/lib/player.py
+++ b/resources/lib/player.py
@@ -57,9 +57,11 @@ class UpNextPlayer(Player):
 
     def onPlayBackEnded(self):  # pylint: disable=invalid-name
         """Will be called when Kodi has ended playing a file"""
-        self.reset_queue()
-        self.api.reset_addon_data()
-        self.state = State()  # Reset state
+        if not self.state.queued:
+            # Reset if playback ended without the next episode being queued
+            self.reset_queue()
+            self.api.reset_addon_data()
+            self.state = State()  # Reset state
 
     def onPlayBackError(self):  # pylint: disable=invalid-name
         """Will be called when when playback stops due to an error"""
--- a/resources/lib/api.py
+++ b/resources/lib/api.py
@@ -51,6 +51,10 @@ class Api:
     def reset_queue():
         jsonrpc(method='Playlist.Remove', id=0, params=dict(playlistid=1, position=0))
 
+    @staticmethod
+    def dequeue_next_item():
+        jsonrpc(method='Playlist.Remove', id=0, params=dict(playlistid=1, position=1))
+
     def get_next_in_playlist(self, position):
         result = jsonrpc(method='Playlist.GetItems', params=dict(
             playlistid=1,
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -145,14 +145,20 @@ class PlaybackManager:
                 elif showing_still_watching_page:
                     still_watching_page.update_progress_control(remaining=remaining, runtime=runtime)
             sleep(100)
-        if next_up_page.is_cancel() and not playlist_item:
+
+        if not playlist_item:
+            self.handle_user_interaction(next_up_page, still_watching_page, episode)
+
+        return showing_next_up_page, showing_still_watching_page
+
+    def handle_user_interaction(self, next_up_page, still_watching_page, episode):
+        if next_up_page.is_cancel():
             # Playback of next episode was cancelled - dequeue the next episode
             self.api.dequeue_next_item()
             self.state.queued = False
-        if still_watching_page.is_still_watching() and not playlist_item:
+        elif still_watching_page.is_still_watching():
             # User is still watching - queue the next episode
             self.state.queued = self.api.queue_next_item(episode)
-        return showing_next_up_page, showing_still_watching_page
 
     def extract_play_info(self, next_up_page, showing_next_up_page, showing_still_watching_page, still_watching_page):
         if showing_next_up_page:
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -50,9 +50,6 @@ class PlaybackManager:
         if not include_play_count or self.state.current_episode_id == episode_id:
             return
 
-        if not playlist_item:
-            self.state.queued = self.api.queue_next_item(episode)
-
         # We have a next up episode choose mode
         if get_setting_int('simpleMode') == 0:
             next_up_page = UpNext('script-upnext-upnext-simple.xml', addon_path(), 'default', '1080i')
@@ -63,7 +60,8 @@ class PlaybackManager:
 
         showing_next_up_page, showing_still_watching_page = self.show_popup_and_wait(episode,
                                                                                      next_up_page,
-                                                                                     still_watching_page)
+                                                                                     still_watching_page,
+                                                                                     playlist_item)
         should_play_default, should_play_non_default = self.extract_play_info(next_up_page,
                                                                               showing_next_up_page,
                                                                               showing_still_watching_page,
@@ -94,7 +92,7 @@ class PlaybackManager:
             # Play local media
             self.api.play_kodi_item(episode)
 
-    def show_popup_and_wait(self, episode, next_up_page, still_watching_page):
+    def show_popup_and_wait(self, episode, next_up_page, still_watching_page, playlist_item):
         try:
             play_time = self.player.getTime()
             total_time = self.player.getTotalTime()
@@ -113,6 +111,9 @@ class PlaybackManager:
         showing_still_watching_page = False
         if int(self.state.played_in_a_row) <= int(played_in_a_row_number):
             self.log('showing next up page as played in a row is %s' % self.state.played_in_a_row, 2)
+            # Queue the next episode
+            if not playlist_item:
+                self.state.queued = self.api.queue_next_item(episode)
             next_up_page.show()
             set_property('service.upnext.dialog', 'true')
             showing_next_up_page = True
@@ -144,6 +145,13 @@ class PlaybackManager:
                 elif showing_still_watching_page:
                     still_watching_page.update_progress_control(remaining=remaining, runtime=runtime)
             sleep(100)
+        if next_up_page.is_cancel() and not playlist_item:
+            # Playback of next episode was cancelled - dequeue the next episode
+            self.api.dequeue_next_item()
+            self.state.queued = False
+        if still_watching_page.is_still_watching() and not playlist_item:
+            # User is still watching - queue the next episode
+            self.state.queued = self.api.queue_next_item(episode)
         return showing_next_up_page, showing_still_watching_page
 
     def extract_play_info(self, next_up_page, showing_next_up_page, showing_still_watching_page, still_watching_page):
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -92,7 +92,7 @@ class PlaybackManager:
             # Play local media
             self.api.play_kodi_item(episode)
 
-    def show_popup_and_wait(self, episode, next_up_page, still_watching_page, playlist_item):
+    def show_popup_and_wait(self, episode, next_up_page, still_watching_page, playlist_item):  # pylint: disable=too-many-branches
         try:
             play_time = self.player.getTime()
             total_time = self.player.getTotalTime()
@@ -145,20 +145,14 @@ class PlaybackManager:
                 elif showing_still_watching_page:
                     still_watching_page.update_progress_control(remaining=remaining, runtime=runtime)
             sleep(100)
-
-        if not playlist_item:
-            self.handle_user_interaction(next_up_page, still_watching_page, episode)
-
-        return showing_next_up_page, showing_still_watching_page
-
-    def handle_user_interaction(self, next_up_page, still_watching_page, episode):
-        if next_up_page.is_cancel():
+        if next_up_page.is_cancel() and not playlist_item:
             # Playback of next episode was cancelled - dequeue the next episode
             self.api.dequeue_next_item()
             self.state.queued = False
-        elif still_watching_page.is_still_watching():
+        if still_watching_page.is_still_watching() and not playlist_item:
             # User is still watching - queue the next episode
             self.state.queued = self.api.queue_next_item(episode)
+        return showing_next_up_page, showing_still_watching_page
 
     def extract_play_info(self, next_up_page, showing_next_up_page, showing_still_watching_page, still_watching_page):
         if showing_next_up_page:


thebertster:
One too many branches for your lint settings in show_popup_and_wait() - would need quite a bit of a refactor to fix that! What do you want to do?

dagwieers:
@thebertster Disable that check. Some of the default checks have weird expectations from developers IMO.

thebertster:
> 
> 
> @thebertster Disable that check. Some of the default checks have weird expectations from developers IMO.

Done and agreed on some of the default pylint checks.

thebertster:
With regards to #203 I think there are quite a few additional things needed to handle user playlists properly. I don't use playlists myself so haven't tested that.

ltGuillaume:
> 
> 
> With regards to #203 I think there are quite a few additional things needed to handle user playlists properly. I don't use playlists myself so haven't tested that.

Using Kodi's videos setting "Play next video automatically" pre-populates the (user) playlist before playback, so in the case I'm describing, it should be straightforward (then again, nothing ever is in software development 😉).

MoojMidge:
> Fixes issues with queuing of next episode via playlist:
> 
> 1. We should not reset state if an episode ends, triggering the onPlayBackEnded event, when we have the next episode queued.

This should happen the other way round. The event is triggered independently of this addon, but while the state should not be reset in this event handler, the queue does need to be reset in the onPlayBackEnded event.

> 2. We should queue the next episode when the "Up Next" dialog is presented, but dequeue the next episode if the user cancels playback of the next episode.

Good pickup. 

> 3. We should not queue the next episode when the "Still Watching?" dialog is presented until and unless the user indicates they are indeed still watching.

Did this cause any problems? Need to be careful to not delay the queuing too much otherwise the original problem from #139 can re-occur.

resources/lib/player.py
```diff
@@ -57,9 +57,11 @@ def onPlayBackStopped(self):  # pylint: disable=invalid-name
...
     def onPlayBackEnded(self):  # pylint: disable=invalid-name
         """Will be called when Kodi has ended playing a file"""
-        self.reset_queue()
-        self.api.reset_addon_data()
-        self.state = State()  # Reset state
+        if not self.state.queued:
+            # Reset if playback ended without the next episode being queued
+            self.reset_queue()
+            self.api.reset_addon_data()
+            self.state = State()  # Reset state
```
MoojMidge, line 64: The queue needs to be reset here, but the state should not be.

See https://github.com/MoojMidge/service.upnext/commit/d5c46a2b9ae2b98898e12d3fa1483915445c49e5 for an alternative method.

thebertster: > 
> 
> > Fixes issues with queuing of next episode via playlist:
> > 
> > 1. We should not reset state if an episode ends, triggering the onPlayBackEnded event, when we have the next episode queued.
> 
> This should happen the other way round. The event is triggered independently of this addon, but while the state should not be reset in this event handler, the queue does need to be reset in the onPlayBackEnded event.

Yes, that makes sense. The problem was with resetting state at that point as it prevents Still Watching? from ever being invoked.
> 
> > 2. We should queue the next episode when the "Up Next" dialog is presented, but dequeue the next episode if the user cancels playback of the next episode.
> 
> Good pickup.
> 
> > 3. We should not queue the next episode when the "Still Watching?" dialog is presented until and unless the user indicates they are indeed still watching.
> 
> Did this cause any problems? Need to be careful to not delay the queuing too much otherwise the original problem from #139 can re-occur.

It's an affirmative action by the user to say they are still watching. While it's true that if the user presses the button at the very last moment before playback ends then the problem from #139 could happen, it's not a major issue from a user experience perspective.

But the alternative would be to enqueue the next episode by default and then dequeue it if the Still Watching dialog times out without the user pressing the button - but the concern there is that a similar race condition could happen where the RPC to dequeue could be processed after the next episode has already started to be played back. Again, from a user experience perspective, this would be a worse scenario than the above.

MoojMidge: > > > Fixes issues with queuing of next episode via playlist:
> > > 
> > > 1. We should not reset state if an episode ends, triggering the onPlayBackEnded event, when we have the next episode queued.
> > 
> > 
> > This should happen the other way round. The event is triggered independently of this addon, but while the state should not be reset in this event handler, the queue does need to be reset in the onPlayBackEnded event.
> 
> Yes, that makes sense. The problem was with resetting state at that point as it prevents Still Watching? from ever being invoked.
> 

I never used the Still Watching? feature, but upon checking my settings I saw it was actually set to 3, so in theory I should have seen the Still Watching? popup. I guess the state reset meant that this never worked on some systems.


> > > 1. We should queue the next episode when the "Up Next" dialog is presented, but dequeue the next episode if the user cancels playback of the next episode.
> > 
> > 
> > Good pickup.
> > > 1. We should not queue the next episode when the "Still Watching?" dialog is presented until and unless the user indicates they are indeed still watching.
> > 
> > 
> > Did this cause any problems? Need to be careful to not delay the queuing too much otherwise the original problem from #139 can re-occur.
> 
> It's an affirmative action by the user to say they are still watching. While it's true that if the user presses the button at the very last moment before playback ends then the problem from #139 could happen, it's not a major issue from a user experience perspective.
> 
> But the alternative would be to enqueue the next episode by default and then dequeue it if the Still Watching dialog times out without the user pressing the button - but the concern there is that a similar race condition could happen where the RPC to dequeue could be processed after the next episode has already started to be played back. Again, from a user experience perspective, this would be a worse scenario than the above.

I'm not sure either scenario is better or worse, one results in failure to playback, the other in unwanted playback. Personally I think it is a bigger issue that the addon doesn't perform its main intended function, but the Still Watching? function should not be broken to achieve this.

There is also a third scenario - where the autoPlayMode has been changed from the default. This could also result in unwanted playback.

I think the only way to resolve this for all scenarios is to use an onAVChange event handler that would stop unwanted playback.

MoojMidge: Actually onAVStarted may not be required.

Can you see if https://github.com/MoojMidge/service.upnext/tree/98f6d65781f18755ab47f3bd6b24bfbbc7470340 works for you?

Should also cover the fourth scenario where unwanted playback could occur - when hitting the close button on the Still Watching? popup.

thebertster: This is not working right. After the nth episode when the "Still Watching?" popup is displayed, I click "Continue watching" and the next episode plays immediately as expected, however UpNext does not trigger for that next episode at all.

MoojMidge: I have seen that happen before but haven't been able to reliably replicate. I believe it is related to the onPlayBackStarted event handler activating (when going to next playlist item) but before there is actually any stream to play.

Should be able to be fixed, but to help narrow down the problem can you clarify:

1. What type of content is being played? Local? Streaming via internet? From local or remote media server?
2. Is there any delay before playback starts once the previous file completes?
3. Observed behaviour is that clicking "Continue Watching" results in the the next episode playing, but after that UpNext fails to activate at all. Does this happen consistently whenever "Continue Watching" is clicked?
4. What happens when "Watch Now" is clicked from a regular UpNext notification, rather than waiting for playback to finish?

thebertster: > * What type of content is being played? Local? Streaming via internet? From local or remote media server?

From a Synology NAS via SMB over gigabit Ethernet and using a shared MariaDb database rather than the local database. Kodi itself is LibreELEC running on a fanless PC. Same issue also seen on a different Kodi running on a Raspberry Pi 3.

> * Is there any delay before playback starts once the previous file completes?

No, there is no perceptable delay.

> * Observed behaviour is that clicking "Continue Watching" results in the the next episode playing, but after that UpNext fails to activate at all. Does this happen consistently whenever "Continue Watching" is clicked?

Yes that's correct - it happens consistently with your build (tested a dozen or so times to be sure it wasn't an intermittent thing).

> * What happens when "Watch Now" is clicked from a regular UpNext notification, rather than waiting for playback to finish?

Just tested this and playback of next episode starts immediately as expected, but again UpNext doesn't trigger for that next episode. Again tried a few times and this is consistent.

I should note that with my original commits on this PR, which I've been using for a week or so now, everything is working fine for all my use cases 100% consistently but as you mention there is also autoPlayNextEpisode and playlists to consider (neither of which I use, so haven't tested behaviour).

I was also thinking about your comment that we should still be calling reset_queue() (but not reset state) in the onPlayBackEnded handler even when self.state.queued==True. I did code that up, and didn't encounter any change in behaviour, but was wondering if it could trigger yet another race condition in some corner case (potentially clearing the playlist *before* Kodi has queued up the next item?). I agree we should call reset_queue() at some point to clear the playlist but I'm thinking that will happen anyway...

Whatever happens subsequently (next episode is successfully played from the playlist, user presses STOP, some error occurs etc.), one of onPlayBackStarted, onPlayBackStopped or onPlayBackError should get triggered, all of which call reset_queue() anyway...

MoojMidge: My setup is DAS via USB3 to Odroid N2 running CoreElec, so faster storage and client somewhere in the middle

It's interesting to get some datapoints to figure out why all these issues are occurring, the original issue from #139 was only an intermittent thing for me, but can still occur depending on how late I click the Still Watching? button with this original PR.

> Just tested this and playback of next episode starts immediately as expected, but again UpNext doesn't trigger for that next episode. Again tried a few times and this is consistent.

I can only replicate this if the 5s delay onPlayBackStarted is reduced, but in doing so have noticed that there are some odd occasions where xbmc.Player().isPlaying() returns True, but xbmc.Player().getTotalTime() returns zero. Waiting for a bit longer results in getTotalTime() returning the correct duration. Perhaps caching has something to do with this? Have you changed the buffermode value to 1 or 4 by any chance? https://kodi.wiki/view/HOW-TO:Modify_the_video_cache#Cache_settings

Can you check if https://github.com/MoojMidge/service.upnext/tree/eacf14278b437720dbe84f77fabab5252ae9a3bf is any better?

> I was also thinking about your comment that we should still be calling reset_queue() (but not reset state) in the onPlayBackEnded handler even when self.state.queued==True. I did code that up, and didn't encounter any change in behaviour, but was wondering if it could trigger yet another race condition in some corner case (potentially clearing the playlist _before_ Kodi has queued up the next item?). I agree we should call reset_queue() at some point to clear the playlist but I'm thinking that will happen anyway...

This possible scenario should be covered, just didn't realise that the state reset was being done incorrectly as I never used the Still Watching? functionality. I'm guessing this was the reason for the issue reported in #135.

I can't trigger this possible scenario normally, but I did test something similar (was trying to see whether removing a playlist item would be enough to prevent the unwanted playback you were experiencing), and removing the playlist item doesn't seem to effect playback. If there was some scenario where the next item had not been queued yet, then the first playlist item gets removed (which does nothing), then the next item may or may not be queued, but reset_queue() should then get called on the next resulting onPlayBackStarted/onAVStarted.

> Whatever happens subsequently (next episode is successfully played from the playlist, user presses STOP, some error occurs etc.), one of onPlayBackStarted, onPlayBackStopped or onPlayBackError should get triggered, all of which call reset_queue() anyway...

The idea behind reset_queue() was to only remove the playlist entry that was just added. If it is only called onPlayBackStopped or onPlayBackError then the playlist will continue to grow. reset_queue() during onPlayBackStarted should cover most normal scenarios, but not also the stop/error events. It would therefore need to be changed to keep track of every item added or clear the entire playlist to cover both scenarios.

MoojMidge: > have noticed that there are some odd occasions where xbmc.Player().isPlaying() returns True, but xbmc.Player().getTotalTime() returns zero. Waiting for a bit longer results in getTotalTime() returning the correct duration

Hmmm this also happens intermittently even after the onAVStarted event fires. It is also possible that this is what is preventing UpNext working with Emby/EmbyCon in some situations.

Another go at fixing the problem: https://github.com/MoojMidge/service.upnext/tree/b8edb738ff0d6fcd897d41de1d19aedb3361aab5

MoojMidge:
Have continued plugging away at fixing this issue and ended up sticking fingers in everything...

https://github.com/MoojMidge/service.upnext/tree/97808a3baea278015a5309f6d5aa95da06e9b245

Not sure the best way to manage this (separate PR?) as the [changes ](https://github.com/im85288/service.upnext/compare/master...MoojMidge:master) are widespread and interconnected, but would appreciate some testing to check if all identified playback issues have been resolved.

Changes include:
- Adds the ability for UpNext to be used as a module to greatly simplify addon integration <sup name="fn1ref">[1](#fn1)</sup>
- Fix issue where Kodi identifies it has started playing a video stream but stream details have not yet loaded
- Fixes issue with seeking to end of file resulting in Kodi playing past the end of the file
- Fixes for handling of scenario where addons provide incorrect notification time/offset
- Fixes multiple timing issues and race conditions resulting from UpNext losing state information when previous streams stop too early/late, new streams start too early/late, video player threads are closed early/late. This should fix #117, #135, #139, #170, and #199. Testing required to confirm. <sup name="fn2ref">[2](#fn2)</sup>
- Addresses unresolved issues from #142
- Fixes issues where changing addon settings can result in UpNext losing state information
- Updated Api methods
- Increased amount of logging
- Allows less restrictive playlist operation. In general if it is in a playlist and playlist handling is enabled, then Up Next will try to handle it. This should address issue identified #203, however the functionality described in [comment-665702929](https://github.com/im85288/service.upnext/issues/203#issuecomment-665702929) may no longer work (if it ever did?)
- Searches for next episode in next season in Kodi library
- Includes implementation of PR #200
- Includes implementation of PR #77


<a name="fn1" href="#fn1ref">[1]</a>: Basic addon integration can be as simple as the following.
In addon.xml:
```
    <requires>
        ...
        <import addon="service.upnext" optional="true"/>
    </requires>
```
    
In code for playing stream:
```
    xbmcplugin.setResolvedUrl(handle, True, listItem)
    
    try:
        import upnext
    except Exception as e:
        xbmc.log(msg='Up Next addon not installed: %s' % e, level=xbmc.LOGWARNING))
        return
    
    upnext_info = dict(
        current_episode=listItem,
        play_url=url
    )
    
    upnext.send_signal(xbmcaddon.Addon().getAddonInfo('id'), upnext_info)
```

<a name="fn2" href="#fn2ref">[2]</a>: I believe this should also resolve a number of other previously reported issues, including #134, #196, #197, but further testing is required.

dagwieers:
@MoojMidge Fantastic!

A PR would definitely be the easiest to review and discuss, however I think the best way forward is to create PR in small stand-alone chunks. For instance, a lot of the bug-fixes will work without any new functionality or restructuring.

I am not saying you need to create PRs for the smallest possible chunk, just a PR for everything that is understandable by itself. (So a bunch of one-line fixes scattered over the code can definitely go together since it makes it possible to review those changes in isolation inside the one PR)

This will also make the process easier to merge the obvious patches asap, and discuss and iterate over the harder/complexer changes in e.g. design.

dagwieers:
For those interested, I think the changes can be viewed here: https://github.com/im85288/service.upnext/compare/master...MoojMidge:master

MoojMidge:
> For instance, a lot of the bug-fixes will work without any new functionality or restructuring.

Maybe. I'll have to have a look at how to split it all up, but the major bug fixes were squashed by moving functionality into UpNextPlayer, from PlayItem, UpNextMonitor, and PlaybackManager, in order to prevent the issues from occurring.

My mistake was also updating State, Api, and utils at the same time, as this was needed to properly keep track of state. Should have done this beforehand.

> however I think the best way forward is to create PR in small stand-alone chunks.

Let me figure out how to use git and I will try and split all this up into manageable PRs

dagwieers:
If it is hard to impossible to get simple fixes out, make one PR and we will see :-)

dagwieers:
PR #77 and PR #200 are definitely not ready to be included though.

dagwieers:
BTW If you make a PR, please create a named branch first, otherwise we might have issues later. `git checkout -b big-rewrite`

MoojMidge:
I can split up some of the less convoluted changes, but there is going to be one big one.

I could try to split it up further but the resulting PRs will likely produce a non-functioning addon when merged, until all related PRs are reviewed/tested/changed/merged. If you could create a test branch I can create the PRs against the test branch to avoid messing up master. Once everything is ok the test branch could be merged back into master.



> PR #77 and PR #200 are definitely not ready to be included though.

With the refactor, implementing #200 was fairly simple, and therefore also easy to remove, but the issues identified in the original PR should no longer be a problem.

The biggest issue I found was that the notification time variables behaved in different ways to each other, and also compared to what their names would suggest (notification_time was a popup duration or offset from end of file, notification_offset was actually a time stamp, and autoPlayCustomTime/autoPlaySeasonTime variables were also popup duration times or offsets from the end of the file).

This was an issue because the effect of #200 is to reduce the effective total file play time by cutting out the credits, which means that variables that were offset from the end of the file ended up being incorrect. Became simple to handle the change in effective run time by making all cue points relative to the start of the file, but needed to rename a bunch of private variables/methods for things to make sense while maintaining compatibility with existing use of the public variables.

#77 is a different case and is something that is now necessary to resolve other bugs. One of the big problem I found was that Kodi does some manner of fast seeking and only seeks to keyframes. When seeking to the end of a file, it doesnt actually seek to the end, but the closest keyframe instead

When this happens programmatically Kodi sometimes (often?) gets confused and thinks that the file has ended and either stops the video too early (creating issues with the Player callbacks) or the video player keeps going (as it should) but then keeps playing past the end of the file (which it shouldn't) and nothing stops it (because Kodi thinks the file had already ended). This was the issue originally reported in #117.

The only way to fix this is to directly play the next file rather than seeking. However directly playing the next file creates other problems - watched state may not be incremented and a resume point may be created depending on popup duration and [Kodi settings](https://kodi.wiki/view/HOW-TO:Modify_automatic_watch_and_resume_points). Therefore #77 would need to be implemented to avoid unintended issues with watch and resume points.

dagwieers:
@MoojMidge I have been test-driving you branch and AFAICT everything has been working fine. I had no specific issues before, but I don't have any now either. :+1:

dagwieers:
@im85288 @thebertster @MoojMidge How we want to proceed with this? If we don't make a plan I fear we will end up (weeks from now) with a stale contribution, or a situation where contributors are unhappy because of wrong expectations or spend effort turned useless. I have been on the receiving end of this, so I can identify ;-)

thebertster:
I was going to test @MoojMidge's build over the weekend.  Looks to be a much more comprehensive approach, so hopefully addresses all these corner cases. Been really manic with work of late so haven't had time. Will let you know the results.

MoojMidge:
@thebertster, @dagwieers There are a few additional bugs that I have seen:
- Kodi can start a new file without onPlayBackEnded triggering (fixed in [master](https://github.com/MoojMidge/service.upnext/commits/master) and [test-instances](https://github.com/MoojMidge/service.upnext/commits/test-instances) branches)
- Popup can get stuck open if new file starts early but then terminates before tracking function can process it (fixed in [test-instances](https://github.com/MoojMidge/service.upnext/commits/test-instances))
- Mistake where I wasn't updating popup progress bar correctly and was using current episode duration rather than next episode duration for end time display on popup (fixed in [test-instances](https://github.com/MoojMidge/service.upnext/commits/test-instances))

If you wouldn't mind testing, could you use [test-instances](https://github.com/MoojMidge/service.upnext/commits/test-instances)? Aside from the changes above, this branch is a step towards allowing for arbitrary class instantiation for the purposes of unit testing.

@dagwieers 
> How we want to proceed with this?

I have tested splitting up the changes and it can be done, but I was hoping a new test branch (copy of current master) could be created here. This way I can create the PRs against this test branch and if anything needs to change, or doesn't work until later PRs are merged, then master will remain unaffected. Let me know what is preferable, either way I can create the PRs (to master or to a new test branch).

thebertster:
@MoojMidge test-instances branch is giving an exception when the popup is displayed (same on my RPi and Intel systems):

```
2020-08-29 15:12:41.669 T:1616180096   ERROR: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                            Error Type: <type 'exceptions.TypeError'>
                                            Error Contents: function takes at least 2 arguments (0 given)
                                              File "/storage/.kodi/addons/service.upnext/resources/lib/service_entry.py", line 10, in <module>
                                                UpNextMonitor().run()
                                              File "/storage/.kodi/addons/service.upnext/resources/lib/monitor.py", line 95, in run
                                                state=self.state
                                              File "/storage/.kodi/addons/service.upnext/resources/lib/playbackmanager.py", line 39, in launch_up_next
                                                play_next, keep_playing = self.launch_popup(episode, playlist_item)
                                              File "/storage/.kodi/addons/service.upnext/resources/lib/playbackmanager.py", line 76, in launch_popup
                                                dialog = UpNextPopup(xmlFilename=filename, item=episode)
                                            TypeError: function takes at least 2 arguments (0 given)
```

MoojMidge:
@thebertster Sorry, forgot to push local changes. Should be up to date now if you wouldn't mind downloading again.

im85288:
Sorry guys for some reason the notifications were going into my spam folder so wasn’t aware of this PR.  @MoojMidge fantastic stuff, really appreciate your input. @dagwieers I always like to encourage contributions so merge when your happy with how things are.

thebertster:
Been using @MoojMidge 's test-instances branch all weekend and not had any issues.

notoco:
I tested on coreelec, on files from the library, as well as on Netflix and on the Netflix + library connection. Works without errors.

dagwieers:
@im85288 @thebertster @notoco @MoojMidge Let us make that PR, and formally review the PR together. I would like to take this opportunity to document as much as possible the logic that is not self-evident so we can improve the readability of this code (which IMO was always a bit lacking before).

As soon as this is merged, I would like to call out to all our stakeholders (and all reporters of open issues) to test the master branch for a few weeks before we can plan to release this.

dagwieers:
Let me close this PR to kick off the new plan ;-)

Thanks everyone for participating!

dagwieers:
@MoojMidge Please don't forget to make a branch before creating your PR. If you need help on Git, let me know.

MoojMidge:
@dagwieers I have started off with some minor PRs.

There will be three major PRs to follow:
PR for [test-merge](https://github.com/MoojMidge/service.upnext/tree/test-merge) - brings things up to date with my master
PR for [test-instances](https://github.com/MoojMidge/service.upnext/tree/test-instances) - master + additional bug fixes + combine popups into one class
PR for [test-monitor](https://github.com/MoojMidge/service.upnext/tree/test-monitor) - master + additional bug fixes + combine popups into one class + monitor performs all monitoring + allow for multiple instances for testing

But I will probably just create a single PR for [test-monitor](https://github.com/MoojMidge/service.upnext/tree/test-monitor) to save reviewing things that have already been changed.

Can discuss further when the minor PRs are reviewed.

PR #206, opened 2020-09-01
Minor update for utils

MoojMidge:
Mostly formatting changes. Only significant change is addition of get_int function

Cherry picked from:
82abfd3a0d109be292bfb345a7a78125259a797a
8b591c7ff0feb57d38b8d0ef93ed117c1f545adb
351265d09b477c83ace44f8b4fdddda4871501a0
37d22f727c309eb5f9a45cee8946d6c2f47ee762
66c8c1f85f6426ba3a788a342cc625ff96675009
7c34cb059bf571c8d2be638151e6d82dcc334c93
41628c5384000604628508fef32d664e27bf244c
024a5d5f87fe1e23c8fa8b699dffbb1914be2ded

### DIFF
--- a/resources/lib/utils.py
+++ b/resources/lib/utils.py
@@ -9,6 +9,7 @@ from xbmcaddon import Addon
 from xbmcgui import Window
 from statichelper import from_unicode, to_unicode
 
+
 ADDON = Addon()
 
 
@@ -47,7 +48,8 @@ def get_setting(key, default=None):
     # We use Addon() here to ensure changes in settings are reflected instantly
     try:
         value = to_unicode(Addon().getSetting(key))
-    except RuntimeError:  # Occurs when the add-on is disabled
+    # Occurs when the add-on is disabled
+    except RuntimeError:
         return default
     if value == '' and default is not None:
         return default
@@ -58,12 +60,14 @@ def get_setting_bool(key, default=None):
     """Get an add-on setting as boolean"""
     try:
         return Addon().getSettingBool(key)
-    except (AttributeError, TypeError):  # On Krypton or older, or when not a boolean
+    # On Krypton or older, or when not a boolean
+    except (AttributeError, TypeError):
         value = get_setting(key, default)
         if value not in ('false', 'true'):
             return default
         return bool(value == 'true')
-    except RuntimeError:  # Occurs when the add-on is disabled
+    # Occurs when the add-on is disabled
+    except RuntimeError:
         return default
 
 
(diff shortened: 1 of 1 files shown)


dagwieers:
So I am not a big fan of just formatting changes. Different people have different styles and we like to keep that out of the changelog and git-blame.

I do not see a good reason to make the proposed changes. I assume you have your editor set to 80 columns, for this project we use 160 columns. Only when you intend to change the specific code anyway, reformatting might be appropriate, but in order to use git blame, keep this to a minimum.

We can discuss changes to formatting separately, e.g. to stick closer to PEP8 or some other formal standard, but only if everyone agrees to it and it improves code readability.

MoojMidge:
> for this project we use 160 columns. Only when you intend to change the specific code anyway, reformatting might be appropriate, but in order to use git blame, keep this to a minimum.

Will do. I'm pretty new to Python so one of the biggest influences in my coding style was whatever flake8 and pylint spat out, which by default included the 79 character line length limit from PEP8.

I can see the benefit in following PEP8, but on the other it seemed far too prescriptive especially when dealing with multiple conditionals with descriptive function/variable names or logging messages.

Didn't really consider use of git blame. You are right - shouldn't change things just for the sake of formatting unless it's something that is agreed upon and consistent throughout the project.

The only significant change here is get_int. Shall I modify the PR?

dagwieers:
If you can modify the PR that is fine, otherwise I will do it myself.

MoojMidge:
One more thing while I am fixing things up related to formatting vs function - I changed the import style in later commits to be
`import foo`
Rather than
`from foo import bar`

I did this because I restructured the modules, and the module namespacing helped improve the code clarity.

Would you and the others involved in this project, be able to comment on this proposed restructure and changed import form, to avoid re-reviewing code that may otherwise need to be reformatted / refactored / restructured?

-  api is a module only, not a class, only contains functions used to interact with Kodi library, player, and playlist, primarily via JSON-RPC. It was originally a class with shared state because it stored addon data.
- dialog contains the UpNextPopup class, single class that handles both types of Up Next notifications.
- monitor contains the UpNextMonitor class, that performs all monitoring, including player monitoring, in order to launch the PlaybackManager when required.
- player contains the UpNextPlayer class, that overrides inbuilt xbmc.Player methods. Monitoring functionality moved to UpNextMonitor, state related variables and methods moved to UpNextState.
- playitem was removed. The library and now playing lookup functionality was incorporated into the api module, state related variables and methods moved to UpNextState without needing to share state between instances.
- playbackmanager contains the PlaybackManager class, that handles the logic for checking for next episodes, showing the Up Next popup, processing user input, and handling playback of the current and next episodes. No longer shares state between instances.
- script is a module for script entry point. Creates a new instance of PlaybackManager, UpNextState, and UpNextPlayer to display a test popup.
- state contains the UpNextState class, that stores all state variables and related methods, including functionality that was previously spread across api, player, playitems, and utils. No longer shares state between instances.
- upnext contains the external addon interface.
- utils contains all miscellaneous helper functions.

dagwieers:
> Would you and the others involved in this project, be able to comment on this proposed restructure and changed import form, to avoid re-reviewing code that may otherwise need to be reformatted / refactored / restructured?

Maybe open a new ticket to discuss this, because I think it got lost in this PR.

I am closing this PR as it only consists of a new function, that is not even used in the PR itself.

PR #207, opened 2020-09-01
Minor fixes

MoojMidge:
Fixes pylint complaints, use monitor rather than sleep

Note that method for creating test popup will be replaced in later PR

### DIFF
--- a/tests/xbmcgui.py
+++ b/tests/xbmcgui.py
@@ -228,14 +228,14 @@ class Window:
 class WindowXML(Window):
     ''' A reimplementation of the xbmcgui WindowXML '''
 
-    def __init__(self):  # pylint: disable=useless-super-delegation
+    def __init__(self):
         ''' A stub constructor for the xbmcgui WindowXML class '''
-        super(WindowXML, self).__init__()
+        super(WindowXML, self).__init__()  # pylint: disable=useless-super-delegation
 
 
 class WindowXMLDialog(WindowXML):
     ''' A reimplementation of the xbmcgui WindowXMLDialog '''
 
-    def __init__(self):  # pylint: disable=useless-super-delegation
+    def __init__(self):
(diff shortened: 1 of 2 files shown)


tests/xbmcgui.py
```diff
@@ -9,7 +9,7 @@
 from xbmcextra import kodi_to_ansi
 
 
-class Control:
+class Control:  # pylint: disable=too-few-public-methods
```
dagwieers, line 12: This is one we prefer to add to .pylintrc, mostly because it is not a sane default IMO.

tests/xbmcgui.py
```diff
@@ -230,12 +230,12 @@ class WindowXML(Window):
 
     def __init__(self):
         ''' A stub constructor for the xbmcgui WindowXML class '''
-        super(WindowXML, self).__init__()
+        super(WindowXML, self).__init__()  # pylint: disable=super-with-arguments
```
dagwieers, line 233: This is a hard one as pylint for py27 is conflicting with pylint for py3.
If we cannot fix that globally, this may be necessary. I would need to check what we do elsewhere.

dagwieers (approved):
I will modify the PR and merge.

tests/xbmcgui.py
michaelarnauts: We've also whitelisted this on our other add-ons.

tests/xbmcgui.py
dagwieers: Ok, this is wrong. Since this is in xbmcgui.py we don't want to do this in .pylintrc, sorry for the mixup.

tests/xbmcgui.py
dagwieers: Ok, this is wrong. Since this is in xbmcgui.py we don't want to do this in .pylintrc, sorry for the mixup.

dagwieers (approved):
All good.

PR #208, opened 2020-09-01
Re-arrange popup module

MoojMidge:
Groundwork for creating upnext module. Move popup classes to new dialog module.

Note that popup classes will be combined in a later PR


resources/skins/default/1080i/script-upnext-stillwatching-simple.xml
```diff
@@ -41,7 +41,7 @@
                         <control type="button" id="10">
                             <label>    $ADDON[service.upnext 30010]</label>
                             <onclick>SendClick(3012)</onclick>
-                            <visible>Integer.IsGreater(Player.TimeRemaining,59)</visible>
+                            <visible>Integer.IsGreater(Window.Property(remaining),59)</visible>
```
dagwieers, line 44: Was this a bug, or is this necessary for something else?

dagwieers (commented):
Not sure why the reorganization. It is a problem for git-blame.
But if it is required for something I am not objecting it.

resources/skins/default/1080i/script-upnext-stillwatching-simple.xml
MoojMidge: It will only become a bug if the functionality in PR#200 is desired to be implemented.

Normally the popup remaining time is the same as Player.TimeRemaing, but not if the popup is meant to do something before the end of the file.

MoojMidge:
> Not sure why the reorganization. It is a problem for git-blame.
> But if it is required for something I am not objecting it.

Two reasons:
- Make room for the addon interface to be in upnext.py for ease of importing into other addons
- There is a lot of duplicate code here that makes the code in playbackmanager difficult to understand and requires more duplicate code paths. Future PR combines both popup classes into one to simplify things

PR #209, opened 2020-09-01
Upnext as module

MoojMidge:
Allows upnext to be imported into other addons to simplify integration.

Basic integration can be as simple as the following

In addon.xml:

```
    <requires>
        ...
        <import addon="service.upnext" optional="true"/>
    </requires>
```
In code for playing stream:

```
    xbmcplugin.setResolvedUrl(handle, True, listItem)
    
    try:
        import upnext
    except Exception as e:
        xbmc.log(msg='Up Next addon not installed: %s' % e, level=xbmc.LOGWARNING))
        return
    
    upnext_info = dict(
        current_episode=listItem,
        play_url=url
    )
    
    upnext.send_signal(xbmcaddon.Addon().getAddonInfo('id'), upnext_info)
```

Nothing changes with the data sent to UpNext with the exception that current_episode and next_episode can now also be a ListItem or VideoInfoTag to avoid having to create a dict to represent the same information.


dagwieers:
I have to think about whether having two interfaces for the same thing is a good idea. Often it is not, unless we want to deprecate the old interface at some point. But I see how this can make things more convenient.

MoojMidge:
The initial reason for this change was because I started adding Up Next integration to a bunch of country specific catchup tv addons, and found myself creating the same file ([from the wiki](https://github.com/im85288/service.upnext/wiki/Example-source-code)) in all the addons and then rewriting the ListItem and/or VideoInfoTag details into a format that Up Next could process.

This creates a number of problems imo:
- It was tedious duplicating and rewriting already existing code
- The only existing interface to Up Next is via Kodi notification, how that notification is actually produced has been left up to the other addons producing the data, as long as the data format matches what Up Next expects.
- If the data required by Up Next ever changes (unlikely) or some other functionality is added/changed/removed, then potentially any addon that relies on this method will need to be updated, whereas having this within Up Next means that the interface exposed via the importable upnext module could remain the same and any changes only need to be made in Up Next

PR #210, opened 2020-09-02
Clean up workflows

dagwieers:
This PR includes:
- Using upstream sonarcloud action with new functionality
- Revert back to using kodi-addon-checker pip package


dagwieers:
@im85288 If we want to have SonarCloud analysis, we need to have SONAR_TOKEN added to the project's secrets. See https://github.com/marketplace/actions/sonarcloud-scan#secrets

@matejmosko I pushed the fix for pylint in this PR otherwise it would not pass our checks. So you can drop it from your PR.

im85288:
@dagwieers cool - added the SONAR_TOKEN secret now

PR #217, opened 2020-10-11
Prepare for v1.1.3 release

### LINKED ISSUE #205: LOGSEVERE and LOGNOTICE will deprecated in kodi/python addons
Please replace with LOGFATAL and LOGINFO. They will be removed in a couple of weeks.

PR: https://github.com/xbmc/xbmc/pull/18346
Forum Post: https://forum.kodi.tv/showthread.php?tid=353818

dagwieers:
Update README and addon.xml for release.

This fixes #205

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,13 @@ The add-on has various settings to fine-tune the experience, however the default
 For [Addon Integration](https://github.com/im85288/service.upnext/wiki/Addon-Integration) and [Skinners](https://github.com/im85288/service.upnext/wiki/Skinners) see the [wiki](https://github.com/im85288/service.upnext/wiki)
 
 ## Releases
+### v1.1.3 (2020-10-14)
+- Enable customPlayTime by default (@dagwieers)
+- Do not seek to the end before playing next episode (@dagwieers)
+- Add Kodi v17 compatibility (@dagwieers)
+- Fix logging on Kodi v19 (Matrix) after recent breakage (@MoojMidge)
+- Enqueue the next episode in the playlist (@MoojMidge)
+
 ### v1.1.2 (2020-06-22)
 - Small bugfix release (@im85288)
 
--- a/addon.xml
+++ b/addon.xml
@@ -1,5 +1,5 @@
 <?xml version='1.0' encoding='UTF-8' standalone='yes'?>
-<addon id="service.upnext" name="Up Next" version="1.1.2" provider-name="im85288">
+<addon id="service.upnext" name="Up Next" version="1.1.3" provider-name="im85288">
     <requires>
         <import addon="xbmc.python" version="2.26.0"/>
     </requires>
@@ -40,6 +40,12 @@ Många befintliga tillägg integreras redan med den här utanför lådan-tjänst
         <platform>all</platform>
         <license>GPL-2.0-only</license>
         <news>
+### v1.1.3 (2020-10-14)
+- Enable customPlayTime by default
+- Do not seek to the end before playing next episode
+- Add Kodi v17 compatibility
+- Fix logging on Kodi v19 (Matrix) after recent breakage
+- Enqueue the next episode in the playlist
 
 ### v1.1.2 (2020-06-22)
 - Bug fix for change from sleep to waitForAbort using seconds and not ms


dagwieers:
cc @im85288 @MoojMidge @michaelarnauts @mediaminister @CastagnaIT Please test this!

mediaminister:
Just found an error on Windows with Kodi 19.

MoojMidge:
Might want to mention that the Still Watching? functionality does not work (as identified in https://github.com/im85288/service.upnext/pull/204#issue-458142906 by @theberster, as a side effect of https://github.com/im85288/service.upnext/pull/198)

dagwieers:
@MoojMidge I would like to add a demo-mode for testing, that when enabled, would play the next video after e.g. 20 seconds of playback. This would allow to easily test StillWatching functionality as well as increase chances of hitting known issues. Which are both now hard to test. Maybe the demo-mode should also reset the resumepoint ;-)

CastagnaIT:
i am try to test (Kodi 19)
I noticed a strange thing when you press the UpNext button to start the next video

when the next video is is about to be started Kodi player call
Player.OnSeek ???
and after
Player.OnResume (this is regular)

OnSeek is a side effect of this upnext version?  I had never noticed with older versions of upnext
I do not understand if the movie is closed there should not be any video seek
or is a new kodi bug introduced?

dagwieers:
I think the OnSeek is to ensure that the episode is marked as watched before closing. Otherwise an episode will appear to not have been finished because the end credits take longer than the margin we have to consider it watched to the end.

CastagnaIT:
I will need to exclude this "fix/option"
this influence the kodi player event notifications and then all addons that use kodi player events to manage his features

in my case this will cause sending irregular data to netflix server

CastagnaIT:
so with netflix if you remenber i have two modes:
- video list with watched status managed by kodi
- video list with watched status managed by addon (values overrided)

the second case this cause the issue

dagwieers:
@CastagnaIT Agreed, it would be even better to mark the episode as been watched using [`SetFileDetails`](https://kodi.wiki/view/JSON-RPC_API/v10#Files.SetFileDetails), except that due to a bug in Kodi this does not work for `plugin://` locations. This is a workaround for the bug IIRC.

michaelarnauts:
> I think the OnSeek is to ensure that the episode is marked as watched before closing. Otherwise an episode will appear to not have been finished because the end credits take longer than the margin we have to consider it watched to the end.
> 
> https://github.com/im85288/service.upnext/blob/6e9d988efc9ab5d5bdc36d954b6886b12989eb9e/resources/lib/playbackmanager.py#L84-L87
> 
> **Update:** Looking at the code I have two remarks here:
> 
> 1. We should probably do this for other methods as well (it is not specific to playlist)
> 2. We could skip it if the offset is sufficiently close to the end (based on [advanced settings](https://kodi.wiki/view/HOW-TO:Modify_automatic_watch_and_resume_points#Settings_explained)

This might be the cause of some issues in VTM Go, the stream starts at -2 (that's actually an issue), so the stream also ends two seconds early. I've experienced some strange crashes, that might be due to seeking to the end while there is no data there?

dagwieers:
I have looked at the history of this `player.seekTime()` statement and this was actually done, not just for marking it as watched, but actually to advance to the next item in the playlist.

https://github.com/im85288/service.upnext/pull/46

This was added since v1.0.2

CastagnaIT:
I would like to try to say one thing
currently the management of watched status (sync) on netflix works as:

show a list of videos with: PlayCount + ResumeTime + TotalTime forced to each listItem

the watched status values, of each video items, are provided by addon itself,
all the values are calculated internally the addon database,
so it must not be forced externally from some jsonrpc

this means that to update all the video watched status is needed only to refresh the page
(so kodi load again the directory and the addon provide the list with watched status updated)

MoojMidge:
> This might be the cause of some issues in VTM Go, the stream starts at -2 (that's actually an issue), so the stream also ends two seconds early. I've experienced some strange crashes, that might be due to seeking to the end while there is no data there?

This is one of many problems that results from seeking towards the end of a stream. It can also cause the VideoPlayer to hang, cause the VideoPlayer to play past the end of the stream or delay the VideoPlayer from closing.

As @dagwieers mentioned the seekTime() call has been in place for some time, however the only benefit it provides over playnext() call is to implicitly cause the playcount and resumepoint to update. I had to remove to resolve a bunch of problems, and instead explicitly update the playcount and resumepoint in the Kodi library, but this obviously only works for local content in  the library.

@CastagnaIT how does the Netflix addon determine whether an item is considered watched or whether to store a resume point, when this is handled directly by the addon operating in sync mode? Does it utilise similar logic as Kodi does internally i.e. as described [here](https://kodi.wiki/view/HOW-TO:Modify_automatic_watch_and_resume_points#Settings_explained)?

dagwieers:
> but this obviously only works for local content in the library.

Not true, it works also for VOD content and can actually update the resumepoints and watched status at the VOD provider as well, as long as the add-on handles `onPlayBackSeek()`, which it should do anyway IMO.

https://github.com/add-ons/plugin.video.vrt.nu/blob/master/resources/lib/playerinfo.py#L120-L130

CastagnaIT:
@MoojMidge yes works something like Kodi:

- if i have the watched status position (provided by the real netflix server) i set as watched when the playback reaches this position
- if i not have the watched status position, then i use the same default kodi behaviour of "playcountminimumpercent" so at 90% of video playback

the resume point is saved while the video is in playingback

MoojMidge:
@dagwieers I was referring to explicitly changing the playcount and resume position via VideoLibrary.SetEpisodeDetails JSON method rather than seeking.

Seeking has the advantage that you don't need to manipulate the episode details, and as you noted, it also works on more than just local content in the library, but it also causes a bunch of other problems due to what appear to be Kodi bugs.

@CastagnaIT, that makes sense. The problem with not seeking to the end of the current file, is that potentially Up Next could play the next file without having reached the 90% point of the current file or reaching the watched status position from Netflix.

Do you see this as a problem for the Netflix addon?

This problem can be resolved for local content in the library, but would need to be handled differently for addons.

At the moment Up Next sends out an NEXTUPWATCHEDSIGNAL notification signal via JSONRPC.NotifyAll after the next file is requested to play (for Trakt integration?). Could the Netflix addon look for this and use it to update the watched status if required?

CastagnaIT:
i my case i use the Player events notifications to send particular http requests to netflix server based on user playback choices (seek play pause stop),
then if you call via code seek event every time, you force me to send request data to the server every time unnecessarily and wrong way

this upnext approach is wrong will produce side effects to all addons that use events notifications to keep track of user playback choices

I am not interested in changing the watched status check metodology it works well as is, this is not the problem

MoojMidge:
@CastagnaIT to clarify I am proposing to remove the seeking in UpNext, not requesting you change how the Netflix watched status was evaluated. However if the seeking was removed there is a potential for another problem to occur and I was trying to understand the consequences of this for the Netflix addon. 

The current code seeks to the end of the file. This creates problems with Kodi, and also potentially for video addons as you describe here:

> i my case i use the Player events notifications to send particular http requests to netflix server based on user playback choices (seek play pause stop),
> then if you call via code seek event every time, you force me to send request data to the server every time unnecessarily and wrong way

While I would like to remove the seekTime() call in UpNext, can you clarify what the actual problem is for the Netflix addon when UpNext seeks to the end of a steam? Is there a difference between UpNext seeking to the end of the stream or a user seeking to the end of the stream? If a video addon is keeping track of the player events to request stream data, shouldn't the seeking by UpNext be handled by the video addon without a problem, in the same way as if it was a user action?

If UpNext did not seek to the end of the stream, then there is the potential for UpNext to activate and play the next stream before the watched point of the current watched point is reached. How would the Netflix addon then keep track of whether the current video was actually watched? Is it not a big issue because the watched point is the same as the end credits time that the Netflix addon sends to UpNext?

dagwieers:
> i my case i use the Player events notifications to send particular http requests to netflix server based on user playback choices (seek play pause stop),
> then if you call via code seek event every time, you force me to send request data to the server every time unnecessarily and wrong way

I do not understand this. When the episode is prematurely ended (because the user decided to go to the next episode) seeking to the end should simply do what a real seek to the end should be doing. I don't see how that would be unnecessary or wrong. It happens only one time.

In what sense is that unnecessary or wrong? What would the behaviour be for you if the users seeks to the end?

CastagnaIT:
>  Is there a difference between UpNext seeking to the end of the stream or a user seeking to the end of the stream?

> If a video addon is keeping track of the player events to request stream data, shouldn't the seeking by UpNext be handled by the video addon without a problem, in the same way as if it was a user action?

yes you can not distinguish if a "seek" is done by code or a seek is done by user interface/remote,
as said before i use these events notifications to send particular http requests to netflix server << with "server" i do not want to say addon service server, but send data to real netflix website server

so you can not accept that all addons can implement stratagems to avoid what you try to impose,
depends in the goal that wants to achieve an addon, and in how a VOD system works

in brief in my case is sent data to the server that say that an user have do an action
this forces the addon not only to send untrue data, but also to spend time/cpu efforts in the service of the addon to do this operation

> How would the Netflix addon then keep track of whether the current video was actually watched?

so the website has no real implementation of the watched state like TRUE/FALSE, takes into account only the elapsed time.
The position that i send to upnext to show the upnext window, is the one that tells kodi that the video has been watched
so this "seek" in this case, is totally useless when i do not use the standard kodi watched status method.

if your goal is to change the watched state of an object, in my opinion:
- suitable tools must be used, if kodi does not provide the appropriate tools like jsonRPC, before you need to make changes to Kodi's core in some way, this would exclude kodi 18, but it will be effective in the next build in a very clean manner, if you plan to use workarounds offer a way to disable it
- the two cases must be taken into account and operate differently: 1) If Listitem has overrided properties so do not change nothing (there is a specific reason if an addon forces these properties and you can't always know), 2) If listitem use normal behaviour you can also change the watched status to ensure that Kodi shows as watched

MoojMidge:
> in brief in my case is sent data to the server that say that an user have do an action
> this forces the addon not only to send untrue data, but also to spend time/cpu efforts in the service of the addon to do this operation

I'm not sure I fully understand this. Conceptually I can't really see a difference between a user seeking/skipping/FF to the end of a stream to quickly skip the end credits, and an addon automating this on behalf of a user as UpNext is doing. I don't think it is untrue data just because the action is triggered by another addon, and if a video addon needs to keep track of current play time to ensure correct stream data is requested from the server, then it seems equally valid to consume resources to do this, regardless of what causes the play time to change.

On the other hand, of course you are right that if it is not necessary, then in the interests of potential optimisation and retaining existing expected behaviour, UpNext should avoid seeking to the end of the file where possible. Removing the seek also eliminates other issues associated with Kodi bugs.

> so the website has no real implementation of the watched state like TRUE/FALSE, takes into account only the elapsed time.
> The position that i send to upnext to show the upnext window, is the one that tells kodi that the video has been watched
> so this "seek" in this case, is totally useless when i do not use the standard kodi watched status method.

Right, so for the Netflix addon, removing the seek will cause little to no issue. This is good. Thanks for taking your time to explain things.

> if your goal is to change the watched state of an object, in my opinion:
> 
> * suitable tools must be used, if kodi does not provide the appropriate tools like jsonRPC, before you need to make changes to Kodi's core in some way, this would exclude kodi 18, but it will be effective in the next build in a very clean manner. if you plan to use workarounds to old version offer a way to disable it
> * the two cases must be taken into account and operate differently: 1) If Listitem has overrided properties so do not change nothing (there is a specific reason if an addon forces these properties and you can't always know), 2) If listitem use normal behaviour you can also change the watched status to ensure that Kodi shows as watched

In my opinion UpNext should not be interacting with ListItem's at all. At the moment it is possible to changed the watched state for library content and that is all that UpNext should be reasonably expected to deal with. Ideally, in the future, the content aggregation/media import concept gets further developed in Kodi and this problem will resolve itself.

As for the v1.1.3 release I would propose:
- Don't seek to end for addon content - same as v1.1.2 behaviour
- Only seek for library content, until the seek can be removed completely with proper workarounds for library content playcount and resume points
- Alternatively just remove the seek now, which could result in potentially false playcounts and resume points for library content. However as this is no different to what currently happens with v1.1.2 for non-playlist content, I don't see removing the seek as a big issue.

dagwieers:
@MoojMidge I do not agree. I do think UpNext should be involved with ensuring an episode that is ended in the credits (so when Up Next was invoked) should ensure that the episode is marked as watched _and_ no longer has any resumepoints stored (either reset to 0 or set to total_time). Not just for local content, but also for VOD content.

Especially in the new implementation where UpNext is starting the next episode itself (not signalling the add-on) there is no means for the add-on to do this (it would not know why the episode was stopped). And as you indicated, there is no reason why an add-on should not be handling this as if the user skips to the end. This should work, if not the add-on has a general issue (and it is not UpNext's fault).

For the issue with Kodi and Kodi's player, I agree we need to prevent any malfunctioning, preferably by fixing Kodi itself. But if we can implement a workaround, that is fine as well.

The problem in the implementation (and I am not sure why this was changed from the original implementation) is that we skip to the end (`player.seekTime()`) _and_ then activate the next episode (`player.playnext()`). I don't know why this was changed, we should simply skip to the end and let Kodi play the next episode in the playlist. Maybe reverting this behaviour fixes the issues we are seeing.

CastagnaIT:
the choice is yours, but force a player Seek only to change watched status is not a good design

if you keep this, you have to mention this peculiarity to the Wiki
because it cause unexpected hidden player event which can cause unexpected behaviours in to the addons

yes nf use "play_info" because don't need the callback anymore
i have also noticed that "play_info" not works in fast way as in it his initial implementation,
now before start next video return to the video list and after play next video

MoojMidge:
@dagwieers In principal I agree UpNext operation should not result in incorrect playcounts or resume positions, however in practice seeking is not the right way to resolve the problem:
- seeking to the end or end-5s is not always possible because Kodi uses fast seeking not exact seeking.
- the interval between keyframes is commonly 2-10s, and the playtime will normally increment past the last keyframe before UpNext tries to seek, which can trigger a bunch of bugs with the Kodi VideoPlayer.
- seeking to the end doesn't always work reliably to update the playcount or reset the resume position.

For library content there is a better way do handle this issue - by using VideoLibrary.SetEpisodeDetails

For addon content, unfortunately at this time there is not much that can be done, but is it actually a significant issue? How many people have actually complained about this?

> i have also noticed that "play_info" not works in fast way as in it his initial implementation,
> now before start next video return to the video list and after play next video

@CastagnaIT does this happen all the time? The play_info method doesn't use the playlist, and is essentially unchanged from previous versions of UpNext.

dagwieers:
> For addon content, unfortunately at this time there is not much that can be done, but is it actually a significant issue? How many people have actually complained about this?

Nobody complained because the implementation works well (this part does). Some add-ons have the problem that they return to the original (unrefreshed) view and there is a bug in Kodi that shows all ListItems with no watched/watching bullets in some cases. But if you refresh the view on return (or return later) this works quite stable.

dagwieers:
>> i have also noticed that "play_info" not works in fast way as in it his initial implementation,
>> now before start next video return to the video list and after play next video

> @CastagnaIT does this happen all the time? The play_info method doesn't use the playlist, and is essentially unchanged from  previous versions of UpNext.

The Netflix add-on is using `play_url` now, no longer `play_info`.

dagwieers:
> - seeking to the end doesn't always work reliably to update the playcount or reset the resume position.

I have not encountered this. So making abstraction of the known implementation bugs, why would it not work reliably? I have a tendency to skip to the end to mark something as watched. Have done it for ages as it is easier than going back and using the context menu. Always worked fine.

dagwieers:
> @dagwieers yes, netflix website server need to know the video events, so every user action, like Start/pause/seek/stop
> to each event are sent some data to the server, to do this i use Player Events
>
> then now Upnext force Skip to end, then what happen now?
> that there are two events when a user choose to play next video:
> The Skip event request and Stop event resquest, that have same data
>
> moltiply this wrong behaviour to 1000+ users and all videos played, make a huge wrong player data to the netflix website server
> then to avoid this behaviour i have to find some strange workaround

Would you consider this to be the wrong behaviour as well when the user skips to the end, and 2 events are being send to Netflix?
Why not?

CastagnaIT:
a case is seek voluntarily at the end done for some reason,
another case is making this behaviour automatic for thousands and more of people creating unnecessary and wrong volumes of data

is not normal because this never happens on the website, nobody jumps at the end in the last seconds, instead a user press "next" button

CastagnaIT:
Github should be include a chat mode for the PR's 😄

CastagnaIT:
ok i think i can try use your "time" condition behaviour to try ignore the Skip event notification

CastagnaIT:
i have do it because i cannot be certain that the credit position provided by netflix is always => 95%

How did you find the "default" 95% value?

dagwieers:
> How did you find the "default" 95% value?

Good question, I seem to be having this wrong?

https://github.com/xbmc/xbmc/blob/master/xbmc/settings/AdvancedSettings.cpp#L144

https://kodi.wiki/view/HOW-TO:Modify_automatic_watch_and_resume_points#Settings_explained

dagwieers:
Sorry guys, for the long discussion, but let's focus on the positive. `seekTime()` is gone now :+1: 

The discussion probably should not have happened in this PR (maybe Slack next time? ;-)), but PR #219 has the conclusive bits.

Let us focus now on testing the master branch. I am confident the user experience is going to be better than ever.

Dnkhatri:
As this plugin does not have a beta test repo could you guys maybe please add an GitHub action to create test versions with a pr. Makes it easier for us non programmers to test. I had to mess around with my fork to generate usable plugins with this pr as the matrix official release has simply become unusable and I could not wait any longer for the update.

dagwieers:
@Dnkhatri Normally you can download the ZIP from the branch and that will work, but because Kodi devs decided to break the backward compatibility in the Add-on metadata, we cannot offer one package for both Matrix and Leia, and we decided to use Leia by default. The change is simple.

What you can do is clone the repository locally and run:

    $ make multizip

This will give you two packages, one for Leia and one for Matrix that will work.

dagwieers:
Can we release this as-is? Have people been testing this?

Apart from the known issues related to automatic playing the next item in the list (when this is unwanted) are there any other issues? Has this been tested on Matrix?

Dnkhatri:
its working fine for me on matrix so far I only use with local files so not sure about streaming addons.  1 problem I had was that the next episode started but the subtitles were stuck from the last episode but I was not able to reproduce it yet so didn't file a bug report

dagwieers:
> 1 problem I had was that the next episode started but the subtitles were stuck from the last episode but I was not able to reproduce it yet so didn't file a bug report

This is not something Up Next is taking care of, so must be a bug in Kodi. It is possible that the use of playlists triggers this issue which was not the case before. If you see this again, please open a new ticket for this to track it.

mediaminister:
> Has this been tested on Matrix?

Yes: https://github.com/im85288/service.upnext/pull/217#issuecomment-707014004

PR #219, opened 2020-10-14
Enable customAutoPlay by default and get rid of seekTime()

dagwieers:
This PR includes:
- Enable `customAutoPlayTime` by default
- No longer skip to end

First of all, skipping the credits is not needed in the majority of cases, because the pop-up appears within the 90% default margin in Kodi.
    
Secondly, this seeking, when done using VOD add-ons, causes the player to exit before the next items is being played, so it doesn't give a nice user experience.

I no longer think this is a real problem, only for videos shorter than 5 mins (or 2.5 mins for customAutoPlay) this becomes an issue.
    
For Netflix it would mean that:
- For a 10 min episode would have issues if the credits shows earlier than 60 seconds.
- For a 30 min episode the credits should not appear before the last 180 seconds.
- For a 1 hour episode the credits should not appear before the last 360 seconds
    
This is all very unlikely. Only for short videos with long credits there might be a risk, but I think that is very theoretical.

### DIFF
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -80,16 +80,8 @@ class PlaybackManager:
         # Signal to trakt previous episode watched
         event(message='NEXTUPWATCHEDSIGNAL', data=dict(episodeid=self.state.current_episode_id), encoding='base64')
         if playlist_item or self.state.queued:
-            # Play playlist media, only seek/skip if media has not already played through
-            if should_play_non_default:
-                try:
-                    total_time = self.player.getTotalTime()
-                    if self.player.getTime() <= total_time * 95 / 100:
-                        #  Seek to near-end to avoid issues with inexact seeking in Kodi Player
-                        self.player.seekTime((total_time - 1) * 59 / 60)
-                except RuntimeError:
-                    pass
-                self.player.playnext()
+            # Play playlist media
+            self.player.playnext()
         elif self.api.has_addon_data():
             # Play add-on media
             self.api.play_addon_item()
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -80,13 +80,16 @@ class PlaybackManager:
         # Signal to trakt previous episode watched
         event(message='NEXTUPWATCHEDSIGNAL', data=dict(episodeid=self.state.current_episode_id), encoding='base64')
         if playlist_item or self.state.queued:
-            try:
-                # Play playlist media, only seek/skip if media has not already played through
-                if should_play_non_default:
-                    self.player.seekTime(self.player.getTotalTime())
-                    self.player.playnext()
-            except RuntimeError:
-                pass
+            # Play playlist media, only seek/skip if media has not already played through
+            if should_play_non_default:
+                try:
+                    total_time = self.player.getTotalTime()
+                    if self.player.getTime() <= total_time * 95 / 100:
+                        #  Seek to near-end to avoid issues with inexact seeking in Kodi Player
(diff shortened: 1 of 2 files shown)


dagwieers:
For a typical 1 hour episode, the 95% margin means 3 minutes. That is far more than Up Next needs to show the pop-up, so that pop-up always falls within the margin that the episode is watched and does not need resumepoints.

im85288 (approved):
Looks good to me

im85288 (approved):

dagwieers:
Also I wonder if it would not be better to pause playback, then seek and then start the next episode. At least you would not jump to the virtual directory and it prevents any potential concurrency conflict in the Kodi Player.

CastagnaIT:
> Also I wonder if it would not be better to pause playback, then seek and then start the next episode. At least you would not jump to the virtual directory and it prevents any potential concurrency conflict in the Kodi Player.

one more event to ignore? I'm starting to get tired of this

dagwieers:
@CastagnaIT No need, I concluded that we don't need any seek() after all. The problem is very marginal.

CastagnaIT:
wait, i call github to say that the Chat implemetation is not more needed🎃

dagwieers:
One big concern with the code is the `should_play_default` and `should_play_non_default`. Looking at the code now I do not understand the complexity that is added and would like to rewrite this part completely. Everytime I read those bits I get overly confused again.

Once we have the demo-functionality added, I plan to rewrite that part completely. Maybe after @MoojMidge's rewrite

im85288:
The should_play_default and should_play_non_default differ depending on the user setting for automatic play of the next episode. It is a bit confusing I agree, but that's what it does in a nutshell. I guess this was one of the problems of having so many config options..it was even worse in the previous app (next up!).

dagwieers:
@im85288 In that case, can you check that it is correct that we _only_ play the next episode on `should_play_non_default` and not on `should_play_default`? Because that is where I am doubting the logic. As this only affects the playlist-method, not the other methods.

im85288:
@dagwieers - so this is only for the playlist method..that part I did not write originally. I am guessing it's a copy/paste issue. I agree for a playlist it's not needed. Playback should always proceed if there is another item to play. So I would say just remove it?

dagwieers:
@MoojMidge @CastagnaIT Based on the default behaviour I think we are good with this simplification. No more `seekTime()`.

Might need more testing on Netflix with short-episode series (10mins and shorter) and choosing next episode as soon as the pop-up appears. I will add a demo-mode to facilitate testing this.

dagwieers:
Let me also add that the 90% is used for concluding that the episode was watched (fully).
But to conclude that a resumepoint is no longer needed, the offset is surprisingly at 92%.

dagwieers:
I have proposed a change in Kodi for this, still requires a change in the documentation if this would get accepted.
https://github.com/xbmc/xbmc/pull/18569

MoojMidge:
> so this is only for the playlist method..that part I did not write originally. I am guessing it's a copy/paste issue. I agree for a playlist it's not needed. Playback should always proceed if there is another item to play. So I would say just remove it?

@im85288, @dagwieers the bit of code in question confused me too at first, but it does work, except when using the Still Watching functionality. In that case `should_play_non_default ` and `should_play_non_default ` are both `False` and even though UpNext doesn't try to play the next video, the playback continues because the next files are already in the playlist. This is an old bug though, and since Still Watching is broken anyway probably not worthwhile trying to fix for v1.1.3

dagwieers:
@MoojMidge Yes, I was not planning to fix this for v1.1.3, unless a fix is trivial . Thanks for confirming this.

PR #220, opened 2020-10-15
Implement DEMO mode

dagwieers:
The Developer section in the settings now has a DEMO mode switch.
If this is enabled, every playback will skip to the near-end to test the
behaviour of the UpNext add-on in high speed.

This allows testing e.g. StillWatching functionality.

### DIFF
--- a/resources/language/resource.language.cs_cz/strings.po
+++ b/resources/language/resource.language.cs_cz/strings.po
@@ -94,6 +94,14 @@ msgctxt "#30049"
 msgid "Next Episode"
 msgstr "Další epizoda"
 
+msgctxt "#30060"
+msgid "[B]UP NEXT DEMO MODE ENABLED[/B]"
+msgstr ""
+
+msgctxt "#30061"
+msgid "[I]Episodes automatically skip to end.[/I]"
+msgstr ""
+
 
 ### SETTINGS
 msgctxt "#30500"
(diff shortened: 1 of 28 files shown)


dagwieers:
cc @im85288 @CastagnaIT @michaelarnauts This should help in debugging UpNext ←→ add-on issues

dagwieers:
The DEMO mode proved useful as it identified already 2 issues in add-ons, and it also made clear why StillWatching logic is broken.

Among others:
- We reset the state after each playback, so also the counter is being reset needlessly
- We queue the next episode in the playlist, but because of this it will always proceed to the next episode (if no action is taken)
- The logic is abstracted into functions at a level where we do not have everything to make the needed decision
- And because of this abstraction the code is hard to comprehend by anyone needing a simple behavioral fix
- And since people have no easy means of testing all the different code paths, the least used functionality got broken

Time to fix it.

dagwieers:
And it is not just StillWatching that is broken, because we have a playlist now with an extra episode, everything that did not require any action before, now will play the next episode (Back, Close, etc.).

It basically plays the next episode in every situation... :-(

dagwieers:
I am going forward and merge this.

This implementation is not completely finished, as it currently only works for add-ons (not for local media) as it is activated when receiving next_episode data from the add-on. Because of this it will only skip to the end when there is actually a next episode, while it should also skip to the end when there is no episode.

PR #221, opened 2020-10-15
Restore the conditional playnext()

dagwieers:
So this served a purpose, now document it better.


dagwieers:
@im85288 So this was important, we don't need to do `playnext()` if the user does not take an action and we proceed to the next episode automatically.

Issue #227, opened 2020-11-06
Action if nothing was chosen dos not work


Yaqwa:
Since last version the next episode starts playing even if I do nothing.
In the settings I said not to play next episode when nothing was chosen. 
This happened since last update from 11/5/2020

psonnosp:
I second that, glad to see that it's not just me having this issue so there's some exclusive error in my setup.

dagwieers:
Yes, we know this. Unfortunately, this is a side-product of using a playlist to advance to the next episode.
The original implementation would not advance to the next episode when it was requested.

siemon-geeroms:
I have a similar issue, when I stop the current episode without choosing an option the next one is being started either way

dagwieers:
@MoojMidge @im85288 Let's discuss the two implementations for future reference: https://github.com/im85288/service.upnext/wiki/Design

And find a better implementation for the next release to fix this issue.

MoojMidge:
On Slack?

The main issue with the method used prior to v1.1.3 (starting playback of next file after end of previous file) is that it stopped working reliably. When the VideoPlayer was exiting, it would do one of three things:

1. Swallow the request to play the next file in the shutdown of VideoPlayer, resulting in the next file not playing.
2. Shutdown the VideoPlayer before starting the next file, triggering the onPlayBackEnded callback prior to the onPlaybackStarted callback of the next file, resulting in UpNext state being lost and Still Watching functionality not working.
3. Open the next file before the VideoPlayer shutdown, resulting in everything working correctly.

The problem with only adding the next file to the playlist near the end of the currently playing file (the just in time method) is that scenarios numbers 1 and 2, above, can still occur.

It was relatively simple to check whether playback should not be occurring, and then explicitly stop playback, to resolve the issue with the playlist method.

siemon-geeroms:
I did some debugging and it appears that when the current file is stopt via api call ( for example with yatse remote) and the up next dialog is already been shown, the "onPlayBackStopped" is not being triggered, hence the queue is not cleared. when I stop the file before the dialog appears it does trigger. I hope this helps.

MoojMidge:
@siemon-geeroms I think that is a separate issue. Are you using Kodi Leia or Kodi Matrix?

siemon-geeroms:
I'm using Kodi leia.

MoojMidge:
[...]

PR #237, opened 2020-12-13
Fix to prevent unwanted playback when using playlist queueing method

MoojMidge:
This is an interim fix to the unwanted playback problem, originally identified by @thebertster in #204 and also subsequently by others in #227.

Note that this wont fix all the issues that were identified in the discussion associated with #204, but should address the most common problem for most installations.

### DIFF
--- a/resources/lib/api.py
+++ b/resources/lib/api.py
@@ -48,7 +48,14 @@ class Api:
         return bool(next_item)
 
     @staticmethod
+    def dequeue_next_item():
+        """Remove unplayed next item from video playlist"""
+        jsonrpc(method='Playlist.Remove', id=0, params=dict(playlistid=1, position=1))
+        return False
+
+    @staticmethod
     def reset_queue():
+        """Remove previously played item from video playlist"""
         jsonrpc(method='Playlist.Remove', id=0, params=dict(playlistid=1, position=0))
 
     def get_next_in_playlist(self, position):
--- a/resources/lib/playbackmanager.py
+++ b/resources/lib/playbackmanager.py
@@ -108,11 +108,14 @@ class PlaybackManager:
         play_item_option_2 = (should_play_non_default and self.state.play_mode == 1)
         if not play_item_option_1 and not play_item_option_2:
             # play_next = False
-            # keep_playing = False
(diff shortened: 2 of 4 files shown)


im85288 (approved):

im85288:
@MoojMidge looks good to me thanks for fixing these issues. I'll await @dagwieers review before merging.

MoojMidge:
No problem.

The only thing to bear in mind with this PR is that the timing issues identified in #204 may now become more apparent now that the playback not working and unwanted playback issues have been worked around.

However, the timing issues probably wont impact as many people as the other problems (I could never directly reproduce the timing issues), and it is a bit too complicated to fix simply/quickly, so on balance, it is probably safe to leave it to later.

Razzeee:
Seems to stop okay now. But it also changed how the `Close` button works, is that expected? It now does, what I would expect from the `Stop` button.

MoojMidge:
Expected? Yes - see https://github.com/im85288/service.upnext/pull/237/files#diff-437f4c1ae1704e3a5689fd1af3f8586471488ab7458c51afc1a157224e33dddcR110-R115

Ideal? No, but this PR was not meant to be a full fix for all the issues that were discovered.

The original code has a couple of issues related to the stop and close buttons and the stopAfterClose setting:
- there is no way to distinguish whether a user pressed the back key, clicked the cancel button, or clicked the stop button. All three actions set the same cancel status variable.
- the Still Watching? dialog does not actually use the stopAfterClose setting, and will never show a stop button.

What should happen (in my opinion) is that:
- pressing the back key should ignore the stopAfterClose setting and only close the open dialog.
- the stopAfterClose setting should function in the Still Watching? dialog and show a stop button rather than a close button, if enabled.
- there should be a separate stop status variable that identifies whether the user clicked the stop button so that immediately stopping playback should only occur when requested by the user.
- if the user does nothing and the default action is set to stop, then playback should also stop immediately.

Didn't want to make more than the bare minimum changes to resolve the original problem, but can put together a bit of a compromise solution.

dagwieers:
I haven't tested this myself, so I am not sure I am the right person to approve this.
But it seems to fix most people's problems so by popular demand...

PR #247, opened 2021-02-10
Fix playlists to work with non-library content

MoojMidge:
Attempt to fix various playlist handling issues as per https://github.com/im85288/service.upnext/issues/134#issuecomment-776325417

Some quick testing indicates that this appears to work as described above, but as the changes were taken from a branch that has diverged significantly from master, probably best if it was tested a little bit more thoroughly.


im85288:
Great stuff will merge into master and hopefully people can test thoroughly

PR #251, opened 2021-03-14
Fix for missing mediatype video info with addon video content

MoojMidge:
Addresses https://github.com/xbmc/xbmc/issues/19378

Video mediatype may be discarded when a listItem VideoInfoTag is merged.

This can prevent UpNext from working with an addon, even if the addon sends the data required for UpNext to function, as UpNext checks the videoplayer.content(episodes) infobool for all video content, when this is only actually needed for library content.

Note that this should also be fixed for playlist content, however that change is a little bit more complicated and requires more significant modifications.

@Dis90 would you mind testing if this resolves the original problem you were having?

### DIFF
--- a/resources/lib/monitor.py
+++ b/resources/lib/monitor.py
@@ -116,3 +116,5 @@ class UpNextMonitor(Monitor):
         self.playback_manager.handle_demo()
         decoded_data.update(id='%s_play_action' % sender.replace('.SIGNAL', ''))
         self.api.addon_data_received(decoded_data, encoding=encoding)
+        self.player.enable_tracking()
+        self.player.reset_queue()
--- a/resources/lib/player.py
+++ b/resources/lib/player.py
@@ -30,6 +30,9 @@ class UpNextPlayer(Player):
     def disable_tracking(self):
         self.state.track = False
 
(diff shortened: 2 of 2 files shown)


Dis90:
I can confirm this fixes original problem what was that Up Next popup didn't show up when watching Up Next started episode if episode was already played in Kodi. I first thought that was happening because listitem mediatype is missing.

im85288:
Great stuff @MoojMidge will merge this in now

MoojMidge:
> I can confirm this fixes original problem what was that Up Next popup didn't show up when watching Up Next started episode if episode was already played in Kodi. I first thought that was happening because listitem mediatype is missing.
> 
> Edit. or was missing mediatype reason why Up Next popup didn't show up?

Yes you were correct, the mediatype was being discarded and UpNext required this to work. However it didn't actually need it, so this PR just removes the requirement for video plugins to set the mediatype, since there are instances where Kodi will discard the information anyway.

The problem is that there is something else going on with how Kodi keeps track of the cached listItem when setResolvedUrl is used, and sometimes the plugin won't even be called, the info merge will never occur, and UpNext will also never be called.

Issue #271, opened 2022-03-10
Request: Ignore Typical DVD/BR Files


CS-Alchemist:
Is it possible to add a feature to ignore typical DVD or BR Files like .iso or .bdmv? I really like this add-on but it interprets a single Video-Element from a BR/DVD as a complete episode and will start the next entry from the library. The only workaround is to deactivate this add-on while playing such files.

MoojMidge:
Can you provide a Kodi log from when you are playing an iso/bdmv when using UpNext?

Are you saying that each individual m2ts is considered as an individual episode when using UpNext? Has the iso/bdmv been scanned into your Kodi library or are you just playing the iso/bdmv from the Kodi file manager?

CS-Alchemist:
Not every m2ts file is considered as an single episode. But the BR playlists are. (mpls files) These files have the playtime information that Kodi uses for the player. When I start the first BR and the first playlist is played, then the Up-Next window comes, just before the end of the first playlist. But this playlist contains only some short video clips, like the animation clip of the distributor, as well as some information about parental control and screening rights, etc. I should note that I always play such blurays with menu and call the index.bdmv for playback.
I would appreciate it very much if with this playback method upnext would generally not react. Because of the BR menu and the binary files on the bluray UpNext is not necessary and Kodi can play the next episode on the disc by itself. The same applies to complete DVD-ISO rips.

Meanwhile, with the new Kodi update the problem disappeared that Upnext at the beginning of bluray playback immediately starts the next entry in the library. (here the same Bluray xD). But the Upnext Messege still appears after a few seconds, counts down its time and does nothing more.

To answer your second question:
Yes all bluray rips and also DVD rips are correctly scanned and stored in the library, as well as all other video files that work with Upnext. This can be seen in the following log.

With the Bluray log it looks quite interesting from line 359.
With the DVD log likewise starting from line 359

[BluRay_Files.log](https://github.com/im85288/service.upnext/files/8758001/BluRay_Files.log)
[DVD_ISO_Files.log](https://github.com/im85288/service.upnext/files/8758003/DVD_ISO_Files.log)

MoojMidge:
Thanks for the log, helps to understand what is happening.

Unfortunately, from the available Python or JSON-RPC APIs, there doesn't seem to be a way to detect that Kodi is playing a disc or disc image. Could check for file types, but your Bluray log indicates that you are using strm files so checking for the file type wouldn't work reliably.

[...]

PR #277, opened 2022-07-26
Update check for filename of multi-part episodes

MoojMidge:
Filenames stored in library may not match filename actually being played e.g. when using strm files or other playlist like files.

Addresses #271


im85288:
@MoojMidge thanks a lot for keeping the development going! Hopefully I can get back to helping out at some stage when my workload calms down

MoojMidge:
No problem. I don't have much time myself, but something like this was fairly simple.

Take it easy :)

MoojMidge:
@im85288 This PR is still in draft, I'll close it and split into separate PRs, rather than lumping the workflow and pylint fixes in with everything else

PR #285, opened 2022-09-12
Release 1.1.8

MoojMidge:
Update addon.xml and README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,18 +20,22 @@ The add-on has various settings to fine-tune the experience, however the default
 
 > NOTE: The add-on settings are found in the Kodi add-ons section, in the *Services* category.
 
-For [Addon Integration](https://github.com/im85288/service.upnext/wiki/Addon-Integration) and [Skinners](https://github.com/im85288/service.upnext/wiki/Skinners) see the [wiki](https://github.com/im85288/service.upnext/wiki)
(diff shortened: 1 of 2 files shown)


im85288:
@MoojMidge thanks for releasing a new version. Looks like the matrix version went in correctly however the leia version failed with:  

"ERROR: service.upnext addon already exists with a lower or equal version: 1.1.5+matrix.1 in nexus branch and the addon has non forward abi compatible dependencies. Users migrating to kodi version nexus won't be able to receive the addon update.
11
ERROR: service.upnext addon already exists with a lower or equal version: 1.1.5+matrix.1 in matrix branch and the addon has non forward abi compatible dependencies. Users migrating to kodi version matrix won't be able to receive the addon update."

Not sure where it got that 1.1.5+matrix.1 version when its target is leia! Anyway maybe we can drop support for leia as most users will be using matrix now.

MoojMidge:
Got there in the end :)

I think the error is not a big deal. It's basically saying that the current version in the matrix/nexus repository (1.1.5+matrix.1) has a lower version number than the version in the leia PR (1.1.8), so someone trying to upgrade Kodi from leia to matrix/nexus won't be able to update the addon and will get a python 2 dependency error.

Once the PR is accepted to the matrix/nexus repository, then the addon-checker should also work for the leia PR because the new version in the matrix/nexus repository (1.1.8+matrix.1) should be higher.

Up to you about leia support, aside from f-strings I don't think there is any major downside to the current backwards compatibility, but I'm not sure what changes nexus requires.

im85288:
Awesome, yeah I wasn't even aware the next version was called nexus. I guess it would be better to assure the addon runs on nexus rather than worry about leia.

Issue #286, opened 2022-10-12
Upnext error ('ascii' codec can't decode)


Bims0n:
Hi, when starting an tvshow episode I get an kodi notification that nextup provoked an error and to check the log. It's kinda hard to reproduce as it doesn't appear to happen on a constant basis and on every single start.
Currently I'm using a custom playlist with the german name format, don't know if the asci error mentioned in the log is related to that. Running the latest upnext version from this git.

The corresponding log can be found here. I think the relevant block is from line 6803 - 6814.
https://paste.kodi.tv/utevidapuk.kodi

MoojMidge:
You are correct, the `Game of Thrones - S04E06 - Die Gesetze von Göttern und Menschen` trips things up in Kodi 18 because of the `ö`.

I think there was an older bug that I made worse recently. Can you check if #287 fixes the problem. Completely untested, and I may have made things even worse, but if you can test it then it should be pretty easy to fix properly.

Bims0n:
Just did a quick test by playing the mentioned GoT episode, looking good now (see screenshot). Log also looks fine. I also did test an episode without any umlaute, still running good as intended. If any future problems occur into that direction I'l keep you updated.

![Unbenannt](https://user-images.githubusercontent.com/111077768/195623778-00406ceb-82a8-4f1a-a8e1-412edc6ec037.PNG)

PR #287, opened 2022-10-13
Fix handling of non-ASCII filenames for Kodi18

MoojMidge:
Possible fix for #286. Completely untested.

Not sure why from_unicode was being used originally?

### DIFF
--- a/resources/lib/monitor.py
+++ b/resources/lib/monitor.py
@@ -6,7 +6,7 @@ from xbmc import Monitor
 from api import Api
 from playbackmanager import PlaybackManager
 from player import UpNextPlayer
-from statichelper import from_unicode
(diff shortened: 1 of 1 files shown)


MoojMidge:
User tested and confirmed problem is fixed.

randomt4sk:
Is there any estimate when this will be released?
This has been fixed more than half a year ago & still hasn't been rolled out.
I'm running a couple of leia clients and manually updating them to nightly would be a quite tedious task.
Or is there a way to directly install the nightly without manually downloading & zipping it?

MoojMidge:
New release is out, but it may take a while to get it into the Kodi addon repository.

Can download the zip file from https://github.com/im85288/service.upnext/releases/tag/v1.1.9

PR #288, opened 2022-11-05
Show next episode regardless of watch status

Jcharrison92:
Is there any way to show the next episode in the widget regardless of watch status? This would be ideal for rewatching series.


MoojMidge:
You have created a pull request to an upstream repository rather than creating an issue.

You can find some info on how to create an issue in the correct place, in these docs: https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-an-issue

PR #290, opened 2023-02-19
Update Swedish translation

Sopor:
Some minor tweaks and corrections

### DIFF
--- a/resources/language/resource.language.sv_se/strings.po
+++ b/resources/language/resource.language.sv_se/strings.po
@@ -7,7 +7,7 @@ msgstr ""
 "Project-Id-Version: service.upnext\n"
 "Report-Msgid-Bugs-To: https://github.com/im85288/service.upnext\n"
 "POT-Creation-Date: YEAR-MO-DA HO:MI+ZONE\n"
-"PO-Revision-Date: 2018-02-26 13:31+0000\n"
+"PO-Revision-Date: 2023-02-19 00:00+0000\n"
 "Last-Translator: Sopor\n"
 "Language-Team: Swedish\n"
 "MIME-Version: 1.0\n"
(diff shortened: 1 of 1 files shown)


resources/language/resource.language.sv_se/strings.po
```diff
@@ -198,16 +198,16 @@ msgstr "Aktivera DEMO-läge"
 
 msgctxt "#30805"
 msgid "Show an UpNext pop-up…"
-msgstr "Visa en popup för UpNext…"
+msgstr "Visa en UpNext-popup…"
```
MoojMidge, line 201: Any reason this is `Visa en UpNext-popup…` and not `Visa en UpNext popup…` or `Visa en UpNext pop-up…`?

resources/language/resource.language.sv_se/strings.po
```diff
@@ -198,16 +198,16 @@ msgstr "Aktivera DEMO-läge"
...
 
 msgctxt "#30807"
 msgid "Show an UpNextSimple pop-up…"
-msgstr "Visa en enkel popup för UpNext…"
+msgstr "Visa en enkel UpNext-popup…"
```
MoojMidge, line 205: Any reason this is `Visa en enkel UpNext-popup…` and not `Visa en enkel UpNext popup…` or `Visa en enkel UpNext pop-up…`?

MoojMidge (commented):
Hi Sopor, thanks for the update.

Just one minor clarification.

resources/language/resource.language.sv_se/strings.po
Sopor: We really don't have any nice translation for pop-up in the swedish language and when i changed it to `UpNext-popup` i can't use two hyphen, so i removed it (you can see that the old translation for pop-up was popup).

When i rearranged the words i need to add a hyphen between the words because UpNext isn't a Swedish word.

If both should have been swedish words i should have spelled it `UpNextpopup` as one word.

PR #292, opened 2023-02-26
Fix SonarCloud integration

MoojMidge:
@im85288 - I noticed that a number of the existing workflows were failing. Have fixed most of them up, but the SonarCloud integration looks like it will need a new user authorisation token to be created.

You will need to go to https://sonarcloud.io/account/security to (re)create the user token.

Then go to https://github.com/im85288/service.upnext/settings/secrets/actions and create a new repository secret named `SONAR_TOKEN` with the secret set to the newly created user token.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,7 @@ jobs:
           -Dsonar.projectKey=im85288_service.upnext
           -Dsonar.sources=resources/lib/
           -Dsonar.tests=tests/
+          -Dsonar.python.coverage.reportPaths=coverage.xml
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}


im85288:
@MoojMidge great stuff and thanks for the detail on fixing the sonar cloud. I will have a look in the next couple of days to fix it, but please ping me if it's not done by next week in case I got distracted.

MoojMidge:
@im85288 - bump

im85288:
@MoojMidge there was already a token in sonarcloud linked, but to be safe I created a new one and updated the value in the secrets section. 

image

Hopefully it's all good now

MoojMidge:
Not sure why it wasn't authenticating before, but that part is now OK.

The analysis execution still fails though - will try and enable logging to figure out why, sometime down the track.

MoojMidge:
Re-ran one of the old CI actions and it failed with an error indicating that automatic analysis was enabled, so the CI action may no longer be necessary. Can be removed if it causes a problem, but the (automatic?) analysis worked anyway.

PR #298, opened 2023-04-15
Fix error when kodi is not playing video any longer 

AnonTester:
Fixes timing issue that sometimes pops up when stopping or forward skipping at end of movie/show just as UpNext is starting up resulting in error:

error <general>: EXCEPTION Thrown (PythonToCppException) : -->Python callback/script returned the following error<--
                                                   Error Type: <class 'RuntimeError'>
                                                   Error Contents: Kodi is not playing any file
                                                     File "/home/kodi/.kodi/addons/service.upnext/resources/lib/service_entry.py", line 8, in <module>
                                                       UpNextMonitor().run()
                                                     File "/home/kodi/.kodi/addons/service.upnext/resources/lib/monitor.py", line 109, in run
                                                       self.playback_manager.launch_up_next()
                                                     File "/home/kodi/.kodi/addons/service.upnext/resources/lib/playbackmanager.py", line 44, in launch_up_next
                                                       episode, source = self.play_item.get_next()
                                                     File "/home/kodi/.kodi/addons/service.upnext/resources/lib/playitem.py", line 61, in get_next
                                                       current_file = self.player.getPlayingFile()
                                                   RuntimeError: Kodi is not playing any file

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,6 @@ For [Addon Integration](https://github.com/im85288/service.upnext/wiki/Integrati
 
 ## Releases
 
-### v1.1.9 (2023-04-15)
-- Fix error when kodi is not playing video any longer
-
 ### v1.1.8 (2022-09-13)
 - Never ask if Still Watching? if playedInARow is 0 (@MoojMidge)
 - Still Watching? checks number of plays not number+1 (@MoojMidge)
--- a/README.md
+++ b/README.md
@@ -24,6 +24,9 @@ For [Addon Integration](https://github.com/im85288/service.upnext/wiki/Integrati
 
 ## Releases
 
+### v1.1.9 (2023-04-15)
+- Fix error when kodi is not playing video any longer
+
 ### v1.1.8 (2022-09-13)
 - Never ask if Still Watching? if playedInARow is 0 (@MoojMidge)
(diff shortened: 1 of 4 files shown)


MoojMidge:
Thanks for this.

Properly handling the `RuntimeError` raised by `xbmc.Player` methods when playback ends is something that could be done in a more general manner, but it is a little bit more involved than using try/catch for each individual method call.

In this particular instance however, catching the exception is probably not quite the right fix. The actual code in `get_next` that triggers the error should probably be changed to use the filename of the currently tracked video instead, as this avoids the potential for Unicode decode errors in Kodi 18, (using Python 2) and also allows the rest of the code to terminate normally if playback has stopped.

https://github.com/im85288/service.upnext/blob/147ed8502fc8a824081db6de5da970548b07fadc/resources/lib/playitem.py#L61

```
    current_file = self.player.get_last_file()
```

Also there is no need to update the changelog and bump the version number, as that is done as part of a separate PR in order to trigger the release workflow.

MoojMidge:
@AnonTester - did you want to update this?

MoojMidge:
Hmm looks like this uncovered a previous old issue - `player.Player` was renamed to `player.UpNextPlayer` but a number of the imports were not updated. Now importing `player.Player` actually imports `xbmc.player` rather than the new class.

Issue #306, opened 2023-10-05
Date format


RentreX:
I am wondering if it is possible to change the date format. 

I am currently using the Arctic Fuse skin and the date is yyyy-mm-dd (yyyy-dd-mm)

Is it possible to change it to dd-mm-yyyy (mm-dd-yyyy)?

I have opened a issue with the dev from Arctic Fuse and they said that it is a add-on issue that they are aware of.

https://github.com/jurialmunkey/skin.arctic.fuse/issues/149#issue-1926567274

![20231005_020910](https://github.com/im85288/service.upnext/assets/60731904/ec990b0f-2361-4e29-8a9d-e74f6aac1564)

MoojMidge:
> it is a add-on issue

Kinda, kinda not.

The date you are seeing is the first aired date of the video (even though it is labelled as the year of the up next video). This date value is not formatted in any way, it is simply whatever date Kodi has stored in its video DB, which by default is stored as a string in the W3C date format of YYYY-MM-DD, and that is what you are seeing.

The skin you are using does not format the value in any way (it could be made to do so, but it is messy and unnecessarily complicated to do), and similarly Up Next does not format the value in way, but could be made to do so (relatively easily)

This is done already here:
https://github.com/MoojMidge/service.upnext/blob/52b38e13739bc23270b17e7390b9bfa46d2fd56c/resources/lib/dialog.py#L173-L181
https://github.com/MoojMidge/service.upnext/blob/52b38e13739bc23270b17e7390b9bfa46d2fd56c/resources/lib/utils.py#L515-L526
 
by formatting the date string to use the same short date format that you have set Kodi to use in your regional settings. Can add the same functionality to Up Next if I have some time over the next month or so.

RentreX:
Not a problem at all.

Thanks for the detailed response.

Yeah I think having it set to the Kodi regional settings would give the end user more flexibility but that is fully up to you to decide.

Thanks again.

MoojMidge:
@RentreX - I'm not sure if you are able to download and apply these changes yourself, but #307 should display the localised date in the popups now. Would be good if you can test it and confirm it works without issue.

RentreX:
@MoojMidge I can confirm that it is working perfectly.

Sorry that the photo is really bad quality. Was way to lazy to get out of bed at 12:30am to turn the light on 😴😂
![20231014_002627](https://github.com/im85288/service.upnext/assets/60731904/d9cbb550-d747-4f1d-a90f-90f410641d55)

PR #307, opened 2023-10-09
Use localised date in popups

MoojMidge:
Refer #306 for details


MoojMidge:
Confirmed working.

SonarCloud check fails as the changes are minimal and the duplicate code is unavoidable.

PR #308, opened 2023-10-31
Update readme & addon.xml description string

wedebe:
This change replaces 'he' as user with generic 'they'.


MoojMidge:
Can you drop the contraction? Seems a bit too conversational.

PR #315, opened 2024-08-09
Finnish translation updates (without the .mo file)

olavinto:
Hopefully this works now. Poedit creates the .mo file automatically and I did not remember that.


olavinto:
Or not... It seems it is still there. God I hate localization via GitHub (vs. Crowdin, Weblate etc... (I do not use it for anything else). :)

PR #316, opened 2024-08-09
Finnish translation updates (without the .mo file)

olavinto:
Thrid time's the charm?


MoojMidge:
Thanks for persisting

